PR #9, opened 2014-08-20
Properly migrate manuals

bwrsandman:
Redo implementation which duplicated manuals before.
Since these are datas, they should not be backed up, the xmlids need to be
renamed.

Depends on the following to be fully OpenUpgrade compatible:
- https://code.launchpad.net/~savoirfairelinux/openupgrade-server/rename_ir_model_data/+merge/231599
- https://code.launchpad.net/~savoirfairelinux/openupgrade-addons/document_page_ir_model_data/+merge/231615

### DIFF
--- a/document_page_health_safety_manual/migrations/7.0.1.0/post-migration.py
+++ b/document_page_health_safety_manual/migrations/7.0.1.0/post-migration.py
@@ -25,8 +25,8 @@ logger = logging.getLogger('upgrade')
 logger.setLevel(logging.DEBUG)
 
 xmlid_renames = [
-    ('document_page_quality_manual.system_quality',
-     'document_page_quality_manual.document_page_quality_manual'),
+    ('document_page_health_safety_manual.wiki_health_safety_manual',
+     'document_page_health_safety_manual.document_page_health_safety_manual'),
 ]
 
 
--- a/document_page_environment_manual/migrations/7.0.1.0/pre-migration.py
+++ b/document_page_environment_manual/migrations/7.0.1.0/pre-migration.py
@@ -22,9 +22,18 @@
 
 import logging
 logger = logging.getLogger('upgrade')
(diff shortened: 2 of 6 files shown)


max3903:
:+1: 

Test and review

pedrobaeza:
It's a huge diff, but repetitive. Seems OK.

:+1: (code review)

document_page_health_safety_manual/migrations/7.0.1.0/post-migration.py
```diff
@@ -33,31 +42,25 @@ def logged_query(cr, query, args=None):
...
-     AND type = 'category';""", (version, category))
+def rename_xmlids(cr, xmlids_spec):
+    """
+    Rename XML IDs. Typically called in the pre script.
+    One usage example is when an ID changes module. In OpenERP 6 for example,
+    a number of res_groups IDs moved to module base from other modules (
+    although they were still being defined in their respective module).
+    """
+    for (old, new) in xmlids_spec:
+        if not old.split('.') or not new.split('.'):
```
gurneyalex, line 53: stylistic remark: it is more efficient to write:

```
if '.' not in old or '.' not in new:
```

as this does not involve paying the price for the splitting and will also stop at the first '.' in the general case.

gurneyalex:
:+1:

document_page_health_safety_manual/migrations/7.0.1.0/post-migration.py
bwrsandman, line 53: This is copy pasted from OpenUpgrade. I don't change it unless the original source is.

bwrsandman, line 53: An even more efficient way would be to split it before the if to avoid spliting twice.

PR #14, opened 2014-11-10
Initialize repo with unported modules and dot files

dreispt:


pedrobaeza:
:+1:

max3903:
I suggest we remove all the modules in **unported**. Modules from 7.0 branch are more up-to-date than the ones from **unported**. We should use the modules from 7.0 branch to migrate them.

dreispt:
Closes #15

max3903:
:+1:

PR #18, opened 2014-11-19
[REF] 8.0 information security manual

llacroix:
Depends on: 
- [x] #23


bwrsandman:
Could you rename `information_security_manual.xml` to `document_page_data.xml` after the model

bwrsandman:
Depends on #23

max3903:
Tested on runbot.

:+1:

information_security_manual/__init__.py
```diff
@@ -0,0 +1 @@
+# -*- encoding: utf-8 -*-
```
dreispt: s/encoding/coding

bwrsandman:
:+1:

PR #16, opened 2014-11-14
[REF] auto-fix conventions

moylop260:

PR #20, opened 2014-11-20
[REF] mgmtsystem

llacroix:
Enabled the module for Installation and ported to the V8.0 API


llacroix:
It seems that the knowledge addons are not installed and for that reasons the build will fail with travis.

pedrobaeza:
You have to add it to `.travis.yml` in order to be recognised. See an example here: https://github.com/OCA/purchase-workflow/blob/8.0/.travis.yml#L23

dreispt:
I restarted the build.

mgmtsystem/mgmtsystem_system.py
```diff
@@ -19,22 +19,15 @@
...
-
-    _defaults = {
-        'company_id': (
-            lambda self, cr, uid, c:
-            self.pool.get('res.users').browse(cr, uid, uid, c).company_id.id),
-    }
+    name = fields.Char('System', size=30, require=True, translate=True)
+    manual = fields.Many2one('document.page', 'Manual')
+    company_id = fields.Many2one('res.company', 'Company',
+                     	  default=lambda self: self.env.user.company_id.id)
```
bwrsandman: This line is not pep8

mgmtsystem/mgmtsystem_system.py
```diff
@@ -19,22 +19,15 @@
...
-        'manual': fields.many2one('document.page', 'Manual'),
-        'company_id': fields.many2one('res.company', 'Company')
-    }
-
-    _defaults = {
-        'company_id': (
-            lambda self, cr, uid, c:
-            self.pool.get('res.users').browse(cr, uid, uid, c).company_id.id),
-    }
+    name = fields.Char('System', size=30, require=True, translate=True)
```
bwrsandman: Is there a strong reason to have a size 30 limit to the name? If not, this parameter should be removed.
Also, did you mean `required` there? It looks like a typo.

mgmtsystem/__openerp__.py
```diff
@@ -48,7 +48,7 @@
         'board_mgmtsystem_view.xml',
     ],
     "demo": [],
-    'installable': False,
+    'installable': True,
     "certificate": ''
 }
 # vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
bwrsandman: Remove the vim line, while you're here.
`certificate` should also be removed.

dreispt:
I think it's best to merge #16 before any ports.
And the dependencies must be ported first, to keep TravisCI happy.

dreispt:
@bwrsandman What if `document_page` dependency was removed? AFAIK it is not strictly necessary, and you might even not want to manage your mgmt-system documents in Odoo.

bwrsandman:
@dreispt I couldn't tell you, @llacroix is the one working on this in v8.
I am still stuck in v7 watching you guys have all the fun :weary:

mgmtsystem/mgmtsystem_system.py
llacroix: I can't really tell why there is a limit size on the name. I fixed the typo.

bwrsandman: In that case, it's probably legacy and would be best to remove it.

llacroix:
I updated the module. I'm not sure we can really remove the document_page dependency. I'm migrating the modules to version 8.0. We haven't discussed of any changes. 

Technically, document_page could be managed outside of mgmtsystem but as I understand they've done it to make it easier to manage documents that are related to management system.

bwrsandman:
@llacroix please rebase your changes with oca/8.0 as the pylint fixes from #16 have been merged and they seem to have created conflicts you code.

They should be minor conflicts

dreispt:
I wouldn't merge this with a red TravisCI, so dependencies should be ported and merged first...

llacroix:
Finally!

mgmtsystem/mgmtsystem_system.py
```diff
@@ -19,22 +19,16 @@
 #
 ##############################################################################
 
-from osv import fields, orm
+from openerp import models, fields
 
+own_company = lambda self: self.env.user.company_id.id
```
pedrobaeza, line 24: I think it will make this static in server init, not calculated with each new record. Can it be?

dreispt, line 24: I think not: `own_company` is a (lambda) function , not an actual value.

dreispt:
:+1:

mgmtsystem/mgmtsystem_system.py
bwrsandman, line 24: Exactly, the static thing is the lambda function. It is not run when imported.

bwrsandman:
100% coverage and no warnings in the log
:+1: 
This won't block, but could you make sure to remove the remaining `#vim` lines

mgmtsystem/mgmtsystem_system.py
```diff
@@ -19,22 +19,16 @@
...
-        'manual': fields.many2one('document.page', 'Manual'),
-        'company_id': fields.many2one('res.company', 'Company')
-    }
-
-    _defaults = {
-        'company_id': (
-            lambda self, cr, uid, c:
-            self.pool.get('res.users').browse(cr, uid, uid, c).company_id.id),
-    }
+    name = fields.Char('System', size=30, required=True, translate=True)
```
bwrsandman: `size=30,` can be removed

mgmtsystem/mgmtsystem_system.py
llacroix, line 24: It could make sense to create a common module to store some lambdas and functions that are used by the other addons. I saw that this `lambda` is present in many places.

pedrobaeza, line 24: Cool to know it. It good be a good addition, but if it is not in the core, it makes no sense to create a module dependency.

PR #21, opened 2014-11-20
Added OCA/knowledge as dependency for travis

llacroix:
It is required for some of the modules that needs to be ported


pedrobaeza:
OK, I fast track it for not blocking other PRs. The idea when I commented in the other PR is to make the modification inside the first PR that depends on this repository, but no matter. We will deal with it :wink:

PR #22, opened 2014-11-20
[REF] 8.0 mgmtsystem_action

llacroix:
The current module inherit `crm.claim` while the `state` attribute doesn't exist anymore. I made some changes to use the `stage_id` instead but I'm not so sure it's a good idea. I'm open to suggestions.


mgmtsystem_action/__openerp__.py
```diff
@@ -43,6 +43,6 @@
         'board_mgmtsystem_action.xml',
     ],
     "demo": ['demo_action.xml'],
-    'installable': False,
+    'installable': True,
 }
 # vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
dreispt: Not a blocker, but it would be nice if you removed the vim lines.

mgmtsystem_action/board_mgmtsystem_action.xml
```diff
@@ -3,11 +3,13 @@
...
 
         <record model="ir.actions.act_window" id="action_mgmtsystem_action">
+            <field name="name">mgmtsystem.action.window</field>
             <field name="res_model">mgmtsystem.action</field>
             <field name="view_type">form</field>
             <field name="view_mode">tree</field>
             <field name="view_id" ref="view_mgmtsystem_action_tree"/>
-            <field name="domain">[('state', 'in', ('draft','open','pending'))]</field>
+	    <!-- TODO check that names match what's present -->
+            <field name="domain">[('stage_id.name', 'in', ('draft','open','pending'))]</field>
```
dreispt: The domain here probably could be removed. if we wnat to keep it, then it should be `[('stage_id.fold','=',False)]`.

mgmtsystem_action/mgmtsystem_action.py
```diff
@@ -77,6 +65,18 @@ def message_auto_subscribe(
             cr, uid, ids, updated_fields, context=context, values=values
         )
 
+    def case_open(self, cr, uid, ids, context=None):
```
dreispt: State was removed remember.
It's best to remove the methods to change state (and corresponding form buttons.)
That's what Odoo did in similar models, such as "Project Issues".

llacroix: What I think I add to add it back because one other module did call thoses methods and they are supposed to be used in the workflow xml file. I wasn't so sure what we should do with that.

mgmtsystem_action/board_mgmtsystem_action.xml
llacroix: ok good

mgmtsystem_action/mgmtsystem_action.py
dreispt: OK. I'm not fond of the workflows, but that's another discussion.

mgmtsystem_action/board_mgmtsystem_action.xml
```diff
@@ -3,11 +3,13 @@
     <data>
 
         <record model="ir.actions.act_window" id="action_mgmtsystem_action">
+            <field name="name">mgmtsystem.action.window</field>
             <field name="res_model">mgmtsystem.action</field>
             <field name="view_type">form</field>
             <field name="view_mode">tree</field>
             <field name="view_id" ref="view_mgmtsystem_action_tree"/>
-            <field name="domain">[('state', 'in', ('draft','open','pending'))]</field>
+	    <!-- TODO check that names match what's present -->
```
bwrsandman: Is this a work in progress?

mgmtsystem_action/mgmtsystem_action.xml
```diff
@@ -50,12 +50,12 @@
             <field name="arch" type="xml">
                 <form string="Action" version="7.0">
                     <header>
-                        <button name="case_cancel" string="Cancel" states="draft,pending" type="object" icon="gtk-cancel"/>
-                        <button name="case_open" string="Open" states="draft,pending" type="object" icon="gtk-go-forward"/>
-                        <button name="case_pending" string="Pending" states="open" type="object" icon="gtk-go-forward"/>
+		        <!-- <button name="case_cancel" string="Cancel" states="draft,pending" type="object" icon="gtk-cancel"/>-->
+		        <!-- <button name="case_open" string="Open" states="draft,pending" type="object" icon="gtk-go-forward"/>-->
+		        <!-- <button name="case_pending" string="Pending" states="open" type="object" icon="gtk-go-forward"/>-->
```
bwrsandman: Better to remove these lines than to comment them out

mgmtsystem_action/board_mgmtsystem_action.xml
llacroix: No, but was looking for some pointers.

mgmtsystem_action/mgmtsystem_action.xml
llacroix: I removed those and some others: filters that don't work anymore.

llacroix:
Can someone have a look why travis crashes?

dreispt:
Restarted it and it doesn't solve the issue.
It seems to be in the `report` module.
Could it be related to https://github.com/odoo/odoo/issues/3811 ?

pedrobaeza:
Yeah, I'm experimenting same issue on l10n-spain branch, so it's related to odoo core. For now, I haven't investigated any solution.

llacroix:
Good news. Seems to work now

dreispt:
The state/stage stuff doesn't bother me a lot, since I believe that should be revisited at  a later moment. So IMO the commented stuff on it can perfectly be left in the code for now.
As long as it's working, you have my :+1:

bwrsandman:
Coverage has gone down -7.69%, please add tests for [mgmtsystem_action/mgmtsystem_action.py](https://coveralls.io/files/365277865) 
- `create`
- `message_auto_subscribe`
- `case_open`
- `case_close`
- `get_action_url`

dreispt:
Come on @bwrsandman: it's not like it's a new module. It's just a port from v7.
We can only demand that it's working on v8 and follows the style guide.

bwrsandman:
This is a strong recommendation, I won't block it, but I won't approve it without tests.

@dreispt I have worked on porting the mgmsystem modules to v7 when I started and I know how much the code needs be better. Tests are the most beneficial when porting to a new version since behaviour is much more likely to do unintended things.

A drop in coverage (esp. from 100%) should be objected.

dreispt:
I see, you're asking it but not demanding - that's OK.
 I would ague the right comparison would be with [v7's](https://github.com/OCA/management-system/tree/7.0) 46%.

From my PoV I would like to see the just "working" in v8 ASAP so that we have the base to improve it. Much of the current code is related to state/stage management, and that will probably be retired.
That's why I don't mind a lot about the missing tests now.

bwrsandman:
@dreispt I don't agree with that sentiment. There will never be a better time to implement tests than now. "Just get it working" is the best strategy to get modules broken in ways only noticeable in production. And since this module is responsible for sending out mass-emails, it is important it works right.

Migration is the only full code revision this module will have. It is the best time to review the whole code. The ideal would be to implement tests in v7 and port those to v8 along with the code to assure there is 1-to-1 same functionality. I am not asking for this, but I would like to see proof that the function is run at some point by travis.

The tests I am asking for are not hard to implement and you might find that some functions are no longer useful and that some are not working as intended.

bwrsandman:
Also, 45fbf8ad589644f7722dec6afb5e573e0df49045 added tests, they are not being run.

I don't know why we're arguing adding tests, when they're there already, but not being run.

@llacroix, there is a problem with your tests, they are not being run.

You need to declare the tests in mgmtsystem_action/tests/__init__.py

``` python
from . import test_create_action

checks = [
    test_create_action,
]
```

mgmtsystem_action/tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+# -*- encoding: utf-8 -*-
+
+from . import test_create_action
```
bwrsandman, line 3: You need to include `test_create_action` in a list called `checks`

llacroix:
Tests are being runned. If they didn't, nothing would call case_open, case_close. After I wrote the tests, coverage went up from 60% to 92%.

mgmtsystem_action/tests/__init__.py
llacroix, line 3: From the documentation, it is only required to import them inside the `__init__` module, the checks might be required to limit tests to certain modules. https://www.odoo.com/documentation/8.0/reference/testing.html

bwrsandman, line 3: Ah, sorry this is new in v8.

bwrsandman:
@llacroix Indeed, I was looking at the Isolated test results. My bad, this is the correct coverage https://coveralls.io/files/365277315

Cases not covered are:
- `case_open()`
  - where `case.stage_id.lower() == 'New'` (I already see a problem here, this will never happen)
  - where `case.user_id == False`
- `case_close()`
  - when there are associated non_conformity_ids (this is wrong as non-conformities are added in another modules)

mgmtsystem_action/mgmtsystem_action.py
```diff
@@ -0,0 +1,107 @@
...
+        return True
+
+    @api.multi
+    def case_close(self):
+        """When Action is closed, post a message on the related NC's chatter"""
+
+        for o in self:
+            if hasattr(o, 'nonconformity_ids'):
+                for nc in o.nonconformity_ids:
+                    nc.case_send_note(_('Action "%s" was closed.' % o.name))
```
bwrsandman: This code should be in the `mgmtsystem_nonconformity` module, it has no place here

mgmtsystem_action/mgmtsystem_action.py
```diff
@@ -0,0 +1,107 @@
...
+        base = super(mgmtsystem_action, self)
+        return base.message_auto_subscribe(updated_fields, values=values)
+
+    @api.multi
+    def case_open(self):
+        """ Opens case """
+
+        for case in self:
+            values = {'active': True}
+            if case.stage_id and case.stage_id.name.lower() == 'New':
```
bwrsandman: This will always evaluate to False as `'New'.lower() == 'new'`

mgmtsystem_action/mgmtsystem_action.py
bwrsandman: Also, the note should have `% o.name` outside of the translation function `_()`

mgmtsystem_action/mgmtsystem_action.py
```diff
@@ -0,0 +1,107 @@
...
+
+own_company = lambda self: self.env.user.company_id.id
+
+
+class mgmtsystem_action(models.Model):
+    _name = "mgmtsystem.action"
+    _description = "Action"
+    _inherit = "crm.claim"
+
+    reference = fields.Char('Reference', size=64, required=True,
```
bwrsandman: Is the size requirement necessary?

mgmtsystem_action/mgmtsystem_action.py
```diff
@@ -0,0 +1,107 @@
...
+        vals.update({
+            'reference': self.env['ir.sequence'].get('mgmtsystem.action')
+        })
+        return super(mgmtsystem_action, self).create(vals)
+
+    @api.multi
+    def message_auto_subscribe(self, updated_fields, values=None):
+        """Automatically add the responsible user to the follow list."""
+        for o in self:
+            self.message_subscribe_users(user_ids=[o.user_id.id],
```
bwrsandman, line 57: `user_id` is not required, so watch out for that case

mgmtsystem_action/mgmtsystem_action.py
```diff
@@ -0,0 +1,107 @@
...
+
+    @api.model
+    def create(self, vals):
+        vals.update({
+            'reference': self.env['ir.sequence'].get('mgmtsystem.action')
+        })
+        return super(mgmtsystem_action, self).create(vals)
+
+    @api.multi
+    def message_auto_subscribe(self, updated_fields, values=None):
```
bwrsandman, line 54: The functionality here looks like something that we should try to merge in odoo 8.0, seems central to `crm.claim` not just actions

mgmtsystem_action/mgmtsystem_action.py
```diff
@@ -0,0 +1,107 @@
...
+            if case.stage_id and case.stage_id.name.lower() == 'New':
+                values['date_open'] = fields.datetime.now()
+
+            if not case.user_id:
+                values['user_id'] = self.env.uid
+
+            values['stage_id'] = self.stage_find(self, None,
+                                                 [('name',
+                                                   '=',
+                                                   'In Progress')])
```
bwrsandman: Only a style suggestion, but this is nicer to format:

```
             values['stage_id'] = self.stage_find(
                 self, None, [('name', '=', 'In Progress')]
             )
```

mgmtsystem_action/mgmtsystem_action.py
```diff
@@ -0,0 +1,107 @@
...
+        config_parameter = self.env['ir.config_parameter']
+        base_url = config_parameter.get_param('web.base_url',
+                                              default='http://localhost:8069')
+
+        query = {'db': self.env.cr.dbname}
+        fragment = {'id': self.id, 'model': self._name}
+
+        return urljoin(base_url, "?%s#%s" % (
+            urlencode(query), urlencode(fragment)
+        ))
```
bwrsandman, line 89: Beyond me, but I think there's a better way to get an url for a record.

Looking at the share module, it seems to do it this way (using `web.base.url` instead of `web.base_url`). 

Maybe there's a way to get a link from `ir.actions.actions`, I don't know.

Anyway, maybe it would be better for the url to come from `share`, maybe not.

mgmtsystem_action/mgmtsystem_action.py
llacroix: From what I understand, if I add this inside the class that inherit this class, it should append in place the methods that are missing. I'll have to test that out, I can probably strip it from there without problem.

mgmtsystem_action/mgmtsystem_action.py
llacroix: Can't tell but the size limit here is strange. The reference is automatically created and it shouldn't be possible to change by hand. It will require to create a huge amount of document to run out of reference ids. Yet that's not impossible. I'll remove it.

mgmtsystem_action/mgmtsystem_action.py
llacroix: I ended up removing both lines, the model doesn't have a `open_date` field.

llacroix:
Should I remove the workflow file, it doesn't seem to be required anymore? I'm not sure it is still relevant.

https://github.com/OCA/management-system/blob/9f8f27efb7e78b8bcf4b91bcfd91df339dbd102e/mgmtsystem_action/workflow_mgmtsystem_action.xml#L60

dreispt:
:clap:

pedrobaeza:
Excellent coverage!

You can remove openupgrade analysis file and migration scripts, because they are for v7. You should build another ones for v8 if there are changes in DB layout.

bwrsandman:
:+1: 
~~You could add a test case with no user_id to get 100% coverage if you want.~~

As for the transition, I still think it's in use. What makes you think it should be removed?

max3903:
Workflow is not relevant if people involved in the action are added as followers and receive notifications of state changes.

pedrobaeza:
@bwrsandman, are you asking me? I ask about removing migration files because it's for v6 to v7 migration, not for v7 to v8.

bwrsandman:
@pedrobaeza Sorry for the confusion: I am asking @llacroix 

> Should I remove the workflow file, it doesn't seem to be required anymore? I'm not sure it is still relevant.

llacroix:
@bwrsandman The workflow transition was calling `case_close` which only do things related to `mgmtsystem_nonconformity` In that case, the workflow for transition should be moved to the `mgmtsystem_nonconformity` module. 

The other workflow for `case_open` seems to send email and the url in the email  is  `http://localhost:8080`. Unless it gets `replaced` automagically somewhere else the url is wrong.

llacroix:
@pedrobaeza  how about updating from 6 to 8 ? Is it possible or the "safe" way would be to update from 6 to 7 and from 7 to 8. 

Right now, there is no migration from 7 to 8 the model didn't change as far as I can tell.

bwrsandman:
@llacroix good point, that should be moved to NC, then,

> Unless it gets replaced automagically somewhere else the url is wrong.

The url will get changed depending on your deployment. Try changing your port, you'll see it changes.

> how about updating from 6 to 8 ?

You can only do an upgrade from one version to the next:
6.0 -> 8.0 is actually 6.0 -> 6.1 -> 7.0 -> 8.0

Maybe in the future OpenUpgrade could support such a high jump but it is not the case today.

That being said, I don't really mind if the old migrations stay, they serve as examples for future migration scripts and they shouldn't cause any problems for people in v8

pedrobaeza:
OK, we have those scripts on v7 branch, but if you want to keep it, it's not a blocking issue.

mgmtsystem_action/mgmtsystem_action.py
llacroix, line 57: Not sure apparently there is a constraint on user_id. User_id can't be null.

llacroix:
Is there anything else to do so we can pass to other modules? Many modules can't be merged because they depend on this one.

dreispt:
@llacroix  To merge this ASAP.
AFAICT there are no blocking issues, right?

bwrsandman:
:+1: No huge issue (in fact it looks amazing), but I would wait for @max3903 to approve

max3903:
:+1:

PR #25, opened 2014-11-21
[REF] 8.0 mgmtsystem nonconformity

llacroix:
Removed base_status dependency and reimplemented some of the methods that were missing to keep the `cases` working. 

depends on
- [x] #22 
- [x] #35
  - [x] #26
    - [x] OCA/knowledge#30


mgmtsystem_nonconformity/mgmtsystem_nonconformity.py
```diff
@@ -22,7 +22,9 @@
 from openerp.tools.translate import _
 from openerp import netsvc
 from openerp.osv import fields, orm
-from openerp.addons.base_status.base_state import base_state
+
+#from openerp.addons.base_status.base_state import base_state
```
bwrsandman: Please remove this commented line

mgmtsystem_nonconformity/mgmtsystem_nonconformity.py
```diff
@@ -411,10 +413,11 @@ def wkf_open(self, cr, uid, ids, context=None):
             )
         self.case_open_send_note(cr, uid, ids, context=context)
         # Open related Actions
-        if o.immediate_action_id and o.immediate_action_id.state == 'draft':
+        # TODO static variables... hmm update state isn't going to work
+        if o.immediate_action_id and o.immediate_action_id.stage_id.name.lower() == 'draft':
             o.immediate_action_id.case_open()
         for a in o.action_ids:
-            if a.state == 'draft':
+            if a.stage_id.name.lower() == 'draft':
```
bwrsandman: This will cause an error in cases where stage_id is not supplied

llacroix: are you sure the stage_id can be null?

mgmtsystem_nonconformity/mgmtsystem_nonconformity.py
```diff
@@ -22,7 +22,9 @@
 from openerp.tools.translate import _
```
bwrsandman: Please split this file so it has one python file by model/class

bwrsandman:
As a general rule, unless the classes are very small (case of one2many). We strongly recommend you split python/xml files by model involved.

https://lists.launchpad.net/openerp-community/msg06955.html

mgmtsystem_nonconformity/mgmtsystem_nonconformity.py
bwrsandman: That doesn't seem to be checked.

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity_cause.py
```diff
@@ -0,0 +1,71 @@
...
+                name = record['parent_id'][1] + ' / ' + name
+            res.append((record['id'], name))
+        return res
+
+    @api.multi
+    def _name_get_fnc(self, prop, unknow_none):
+        res = self.name_get()
+        return dict(res)
+
+    _constraints = [
```
llacroix: _constraint not present in mgmtsystem_nonconformity_origin.py. yet the code is almost the same. We could technically subclass both class and clone the implementation or just add a type instead of copying code.

bwrsandman: Good catch on the constraint.

I haven't done this, but I think to clone a class you can do this:

``` python
class mgmtsystem_nonconformity_origin(models.Model):
    _name = "mgmtsystem.nonconformity.origin"
    _inherit = "mgmtsystem.nonconformity.cause"
```

If that doesn't work, you can use an abstract model and then subclass from that

llacroix: Yes, but I'd rather subclass them as origin aren't causes and causes aren't origin. If it's fine to inherit only for the structure only then it could be fine but if there is a way to get all origins and it will catch all causes too it won't make sense.

bwrsandman:
depends on ~~#35~~

bwrsandman:
pep8:

```
./mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py:79:8: E121 continuation line under-indented for hanging indent
```

dreispt:
Could we have this to not depend on `document_page_procedure`?
We may want to manage NCs is Odoo but prefer to manage Procudure Docs elsewhere.

max3903:
@dreispt having a dependency on document_page_procedure does not necessarily mean you manage your documentation content in Odoo. We use it as a list to reference the content in another application. The day we want to put everything in Odoo, we just have to transfer the content and drop the other application.

If you want to drop the dependency, you will have to create a new field to provide the list of procedures (quality), environmental aspects (environment), and other thing in other systems. 

I am sure you will agree with me that any Management Systems software should come with a documentation system, so why Odoo should be different here.

mgmtsystem_nonconformity/board_mgmtsystem_nonconformity.xml
```diff
@@ -3,6 +3,7 @@
     <data>
 
         <record id="action_mgmtsystem_nonconformity" model="ir.actions.act_window">
+            <field name="name">board.mgmtsystem.nonconformity.act_window</field>
```
dreispt: name should be a human readable text

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -151,8 +42,12 @@ class mgmtsystem_nonconformity_severity(orm.Model):
 ]
 _STATES_DICT = dict(_STATES)
 
+own_company = lambda self: self.env.user.company_id.id
+default_date = lambda *a: time.strftime(DATE_FORMAT)
+default_user_id = lambda self: self.env.user.id
```
dreispt: I bet assigning lambdas doesn't pass lint checks...

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -162,127 +57,129 @@ class mgmtsystem_nonconformity(base_state, orm.Model):
...
-            self.pool.get('res.users').browse(cr, uid, uid, c).company_id.id),
-        'date': lambda *a: time.strftime(DATE_FORMAT),
-        'state': 'draft',
-        'author_user_id': lambda cr, uid, id, c={}: id,
-        'ref': 'NEW',
-    }
-
-    def create(self, cr, uid, vals, context=None):
+    # 1. Description
+    id = fields.Integer('ID', readonly=True)
```
dreispt: I think adding `id` is not needed anymore with the new API. Can you confirm?

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -162,127 +57,129 @@ class mgmtsystem_nonconformity(base_state, orm.Model):
...
+    ref = fields.Char(
+        'Reference',
+        size=64,
+        required=True,
+        readonly=True,
+        default="NEW"
+    )
+    date = fields.Date('Date', required=True, default=default_date)
```
dreispt: I would rather have lambdas inline

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -162,127 +57,129 @@ class mgmtsystem_nonconformity(base_state, orm.Model):
...
+    partner_id = fields.Many2one('res.partner', 'Partner', required=True)
+    reference = fields.Char('Related to', size=50)
```
dreispt: drop size attribute?

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -162,127 +57,129 @@ class mgmtsystem_nonconformity(base_state, orm.Model):
...
         vals.update({
-            'ref': self.pool.get('ir.sequence').get(
-                cr, uid, 'mgmtsystem.nonconformity')
+            'ref': self.env['ir.sequence'].get('mgmtsystem.nonconformity')
         })
-        return super(mgmtsystem_nonconformity, self).create(
-            cr, uid, vals, context)
+        return super(mgmtsystem_nonconformity, self).create(vals)
 
     def message_auto_subscribe(
```
dreispt: The messaging features need a rewrite. I'm not bothered by the direct port, but wouldn't mind if they were removed now.

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -447,14 +346,17 @@ def wkf_cancel(self, cr, uid, ids, context=None):
     def wkf_close(self, cr, uid, ids, context=None):
         """Change state from in progress to closed"""
         o = self.browse(cr, uid, ids, context=context)[0]
-        done_states = ['done', 'cancelled']
+        # TODO make it more friendly
+        done_states = ['done', 'cancelled', 'settled', 'rejected']
```
dreispt: settled and rejected are not in the states list defined at the top

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -447,14 +346,17 @@ def wkf_cancel(self, cr, uid, ids, context=None):
...
         if (o.immediate_action_id
-                and o.immediate_action_id.state not in done_states):
+                and o.immediate_action_id.stage_id.name.lower()
```
dreispt: This is mixing up sates with states!
Something is not right here.

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity_cause.py
```diff
@@ -0,0 +1,71 @@
...
+
+class mgmtsystem_nonconformity_cause(models.Model):
+    """
+    Cause of the nonconformity of the management system
+    """
+    _name = "mgmtsystem.nonconformity.cause"
+    _description = "Cause of the nonconformity of the management system"
+    _order = 'parent_id, sequence'
+
+    id = fields.Integer('ID', readonly=True)
```
dreispt: confirm if explicit id can be removed

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity_cause.py
```diff
@@ -0,0 +1,71 @@
...
+    child_ids = fields.One2many(
+        'mgmtsystem.nonconformity.cause',
+        'parent_id',
+        'Child Causes',
+    )
+    ref_code = fields.Char('Reference Code')
+
+    @api.multi
+    def name_get(self):
+        reads = self.read(['name', 'parent_id'])
```
dreispt: Don't use read. Iterate over the `self` recordset directly.

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity_cause.py
```diff
@@ -0,0 +1,71 @@
...
+from openerp.osv import osv
+
+
+class mgmtsystem_nonconformity_cause(models.Model):
+    """
+    Cause of the nonconformity of the management system
+    """
+    _name = "mgmtsystem.nonconformity.cause"
+    _description = "Cause of the nonconformity of the management system"
+    _order = 'parent_id, sequence'
```
dreispt, line 31: We should active hierarchical relations setting `_parent_store=True`. See https://www.odoo.com/documentation/8.0/reference/orm.html#openerp.models.Model._parent_store

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -21,8 +21,9 @@
 
 from openerp.tools.translate import _
 from openerp import netsvc
-from openerp.osv import fields, orm
-from openerp.addons.base_status.base_state import base_state
+from openerp.osv import orm
```
dreispt: You shouldn't need this now...

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity_origin.py
```diff
@@ -0,0 +1,65 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp import fields, models, api
+
+
+class mgmtsystem_nonconformity_origin(models.Model):
```
dreispt: Same comments as for the `cause` model

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity_origin.py
```diff
@@ -0,0 +1,65 @@
...
+            name = record['name']
+            if record['parent_id']:
+                name = record['parent_id'][1] + ' / ' + name
+            res.append((record['id'], name))
+        return res
+
+    @api.multi
+    def _name_get_fnc(self, prop, unknow_none):
+        res = self.name_get()
+        return dict(res)
```
dreispt: I don't think this does anything useful.

mgmtsystem_nonconformity/tests/test_origin.py
```diff
@@ -0,0 +1,35 @@
...
+        record = self.env['mgmtsystem.nonconformity.origin'].create({
+            "name": "TestOrigin",
+        })
+
+        name_assoc = record.name_get()
+        name_dict = record._name_get_fnc(None, None)
+
+        self.assertEqual(name_dict[record.id], "TestOrigin")
+        self.assertEqual(name_assoc[0][1], "TestOrigin")
+        self.assertEqual(name_assoc[0][0], record.id)
```
dreispt, line 48: Kudos for the tests :trophy:

mgmtsystem_nonconformity/views/mgmtsystem_cause.xml
```diff
@@ -0,0 +1,62 @@
...
+    <data>
+        <!-- Cause -->
+
+        <record id="view_mgmtsystem_nonconformity_cause_form" model="ir.ui.view">
+            <field name="name">mgmtsystem.nonconformity.cause.form</field>
+            <field name="type">form</field>
+            <field name="model">mgmtsystem.nonconformity.cause</field>
+            <field name="arch" type="xml">
+                <form string="Causes of nonconformity">
+                <field name="name"/>
```
dreispt: I think you need a `<group>` element around the fields for the labels to be displayed.

mgmtsystem_nonconformity/views/mgmtsystem_nonconformity.xml
```diff
@@ -80,7 +80,7 @@
                         <button name="button_close" string="Close" class="oe_highlight"
                             states="open" groups="mgmtsystem.group_mgmtsystem_manager,mgmtsystem.group_mgmtsystem_auditor"/>
                         <button name="case_reset" string="Reset to New" states="done,cancel" type="object"/>
-                        <field name="state" widget="statusbar" statusbar_visible="draft,analysis,pending,open,done"/>
+                        <field name="state" widget="statusbar" statusbar_visible="draft,analysis,pending,open,done" clickable="True"/>
```
dreispt: I don't think you want this to be "clickable": flow needs to be controlled by logic in buttons

mgmtsystem_nonconformity/views/mgmtsystem_origin.xml
```diff
@@ -0,0 +1,63 @@
...
+
+        <!-- Origin -->
+        
+        <record id="view_mgmtsystem_nonconformity_origin_form" model="ir.ui.view">
+            <field name="name">mgmtsystem.nonconformity.origin.form</field>
+            <field name="type">form</field>
+            <field name="model">mgmtsystem.nonconformity.origin</field>
+            <field name="arch" type="xml">
+                <form string="Origins of nonconformity">
+                    <field name="name"/>
```
dreispt: add `<group>` element

mgmtsystem_nonconformity/views/mgmtsystem_severity.xml
```diff
@@ -0,0 +1,32 @@
...
+    <data>
+        <!-- Severity -->
+
+        <record id="view_mgmtsystem_nonconformity_severity_form" model="ir.ui.view">
+            <field name="name">mgmtsystem.nonconformity.severity.form</field>
+            <field name="type">form</field>
+            <field name="model">mgmtsystem.nonconformity.severity</field>
+            <field name="arch" type="xml">
+                <form string="Severity of Complaints and Nonconformities">
+                        <field name="name"/>
```
dreispt: add `<group>` element

mgmtsystem_nonconformity/views/mgmtsystem_cause.xml
llacroix: Good point

max3903:
Needs rebase, I think.

llacroix:
Working on it now

dreispt:
@llacroix You are still to address my comments, right?

llacroix:
@dreispt yes I'm on it... I'll push updates

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
llacroix: It shouldn't but the travis tests did pass. There might be something wrong with the tests.

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
llacroix: Yes, as far as I can say, the id should be created by default even on openerp 7 as it is used in ir_model_data. Not sure how it worked before but it shouldn't be necessary. I'll test it thought.

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
llacroix: This is a problem, I wasn't sure how to handle this. The problem come from the fact that those states are on a `mgmtsystem.action` object. The `mgmtsystem.action` object is a `crm.claim object`. But the `crm.claim` object doesn't have states anymore but stages.

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
llacroix: The object mgmtsystem.action doesn't have states anymore. I can add them back and ignore stages.

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
llacroix: Changed form `from openerp.exceptions import except_orm`

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity_origin.py
llacroix: Removed that part

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
llacroix: Removed

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
llacroix: Done, yet having named function makes it reusable. if we had a common module. The `own_company` is used in other modules but each copy the same method (lambda)

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
llacroix: Done

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
llacroix: You'll have to be more precise if you want something to get removed.

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity_cause.py
llacroix: Done

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity_cause.py
llacroix, line 31: we only have to add `_parent_store = True` ?

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity_origin.py
llacroix: Not sure to which comment you're referencing to.

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
dreispt: No, please keep Stages.

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity_cause.py
dreispt, line 31: No, two helper fields also need to be added:

```
_parent_store = True
parent_left = fields.Integer('Parent Left', index=True)
parent_right = fields.Integer('Parent Right', index=True)
```

llacroix, line 31: It finds by itself how to construct the left and right? I understand that it creates a structure that will allow fetching part of the trees efficiently but the doc doesn't tell much how it finds the left and right. I'll add it and lets see how it goes.

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity_origin.py
llacroix: @dreispt it's about the parent_id right?

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity_cause.py
llacroix, line 31: Added the fields.

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity_origin.py
llacroix: Added the fields.

llacroix:
After looking at the current modules, I believe we might have to change `mgmtsystem.action` to use states instead. It might not be necessary to make it inherit from `crm.claim`

mgmtsystem_nonconformity/tests/test_nonconformity.py
```diff
@@ -0,0 +1,44 @@
+from openerp.tests import common
```
bwrsandman: Could you test the workflow too?

llacroix: will do.

llacroix:
https://coveralls.io/files/678256153

bwrsandman:
@llacroix https://coveralls.io/builds/2466939/source?filename=mgmtsystem_nonconformity%2Fmodels%2Fmgmtsystem_nonconformity.py

mgmtsystem_nonconformity/tests/test_workflows.py
```diff
@@ -0,0 +1,222 @@
...
+                conf_obj.action_sign_analysis()
+
+        self.try_signal(conf_obj, 'button_review_n', 'pending')
+        self.assertEqual(conf_obj._state_name()[conf_obj.id],
+                         'Pending Approval')
+        self.try_cancel(conf_obj)
+
+        # Test Pending to Open
+        # pending -> analysis
+        # self.try_invalid_signal(conf_obj, 'button_analysis_p', 'analysis')
```
bwrsandman: Dead code?

mgmtsystem_nonconformity/tests/test_workflows.py
```diff
@@ -0,0 +1,222 @@
...
+        self.assertEqual(conf_obj._state_name()[conf_obj.id], 'Analysis')
+        self.try_cancel(conf_obj)
+
+        # Test Analysis to pending
+        # pending -> analysis (already on analysis)
+        self.try_signal(conf_obj, 'button_analysis_p', 'analysis')
+        # draft-> analysis (already on analysis)
+        self.try_signal(conf_obj, 'button_analysis_n', 'analysis')
+        # analysis -> pending
+        # self.try_invalid_signal(conf_obj, 'button_review_n', 'pending')
```
bwrsandman: Dead code?

mgmtsystem_nonconformity/tests/test_workflows.py
```diff
@@ -0,0 +1,222 @@
...
+        # analysis -> pending
+        self.try_invalid_signal(conf_obj, 'button_review_n', 'pending')
+        # open -> pending
+        self.try_invalid_signal(conf_obj, 'button_review_p', 'pending')
+        # pending-> open
+        self.try_invalid_signal(conf_obj, 'button_open', 'open')
+        # open -> done
+        self.try_invalid_signal(conf_obj, 'button_close', 'done')
+        # draft-> analysis
+        # self.try_invalid_signal(conf_obj, 'button_analysis_n', 'analysis')
```
bwrsandman: Dead code?

mgmtsystem_nonconformity/tests/test_workflows.py
```diff
@@ -0,0 +1,222 @@
...
+            sid = self.cr.savepoint()
+            with sid:
+                obj.signal_workflow(signal_name)
+                obj.refresh()
+                self.assertEqual(obj.state, state)
+
+                if rollback:
+                    raise BreakSavePoint()
+        except BreakSavePoint:
+            obj.refresh()
```
bwrsandman: Why is it necessary to have a custom error? Is it because of the context manager? Maybe you could define the Exception within this function in order to keep it encapsulated.

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -0,0 +1,425 @@
...
+
+    def _state_name(self):
+        res = dict()
+        for o in self:
+            res[o.id] = _STATES_DICT.get(o.state, o.state)
+        return res
+
+    name = fields.Char('Name')
+
+    # 1. Description
```
bwrsandman: This could be put in the helptext

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -0,0 +1,425 @@
...
+]
+_STATES_DICT = dict(_STATES)
+
+
+class mgmtsystem_nonconformity(models.Model):
+
+    """Management System - Nonconformity."""
+
+    _name = "mgmtsystem.nonconformity"
+    _description = "Nonconformity of the management system"
```
bwrsandman: Nonconformity of the Management System

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -0,0 +1,425 @@
...
+    ('open', _('In Progress')),
+    ('done', _('Closed')),
+    ('cancel', _('Cancelled')),
+]
+_STATES_DICT = dict(_STATES)
+
+
+class mgmtsystem_nonconformity(models.Model):
+
+    """Management System - Nonconformity."""
```
bwrsandman: Docstring is redundant, if it doesn't explain what it is, it can be removed

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -0,0 +1,425 @@
...
+
+    def case_send_note(self, cr, uid, ids, text, data=None, context=None):
+        for id in ids:
+            pre = self.case_get_note_msg_prefix(cr, uid, id, context=context)
+            msg = '%s <b>%s</b>' % (pre, text)
+            if data:
+                o = self.browse(cr, uid, ids, context=context)[0]
+                post = _(u'''
+<br />
+<ul><li> <b>State:</b> %s → %s</li></ul>\
```
bwrsandman: Couldn't this be done simply by defining the track on the state column definition?

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -0,0 +1,425 @@
...
+''') % (o.state, data['state'])
+                msg += post
+            self.message_post(cr, uid, [id], body=msg, context=context)
+        return True
+
+    def case_get_note_msg_prefix(self, cr, uid, id, context=None):
```
bwrsandman: `@api.model`

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -0,0 +1,425 @@
...
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+from openerp.tools.translate import _
+from openerp import netsvc
+from openerp.exceptions import except_orm
```
bwrsandman: These `except_orm` could be replaced by more specific exceptions
`excpetions` can be added to the `from openerp` line

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -0,0 +1,425 @@
...
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+from openerp.tools.translate import _
+from openerp import netsvc
```
bwrsandman: Could be added to line 24

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -0,0 +1,425 @@
...
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+from openerp.tools.translate import _
```
bwrsandman: You can import `_` directly `from openerp` on line 24

mgmtsystem_nonconformity/models/mgmtsystem_action.py
```diff
@@ -0,0 +1,42 @@
...
+    nonconformity_ids = fields.Many2many(
+        'mgmtsystem.nonconformity',
+        'mgmtsystem_nonconformity_action_rel',
+        'action_id',
+        'nonconformity_id',
+        'Nonconformities',
+        readonly=True,
+    )
+
+    # TODO add case_open/case_close conditions
```
bwrsandman: Can an issue be added for this?

mgmtsystem_nonconformity/board_mgmtsystem_nonconformity.xml
bwrsandman: Or it can be ommited

mgmtsystem_action/tests/test_create_action.py
```diff
@@ -11,7 +11,8 @@ def test_create_action(self):
         self.assertEqual(record.name, "SampleAction")
         self.assertNotEqual(record.reference, "NEW")
         self.assertEqual(record.type_action, "immediate")
-        self.assertEqual(record.stage_id.name, "New")
+        self.assertEqual(record.stage_id.name, "Draft")
```
bwrsandman: The use of `_()` would be more correct if it works.

dreispt:
Current status is needs fixing, correct?

bwrsandman:
@dreispt Yes, but unfortunately, @llacroix won't be working on it anymore.

bwrsandman:
@max3903, @dreispt please test

dreispt:
Thanks @bwrsandman !

mgmtsystem_action/mgmtsystem_action.py
```diff
@@ -45,6 +45,18 @@ class mgmtsystem_action(models.Model):
     system_id = fields.Many2one('mgmtsystem.system', 'System')
     company_id = fields.Many2one('res.company', 'System', default=own_company)
 
+    stage_id = fields.Many2one(
+        'mgmtsystem.action.stage',
+        'Stage',
+        default=lambda self: self.get_default_stage()
```
dreispt, line 51: You can use: `default=get_default_stage`

mgmtsystem_action/mgmtsystem_action.py
```diff
@@ -45,6 +45,18 @@ class mgmtsystem_action(models.Model):
...
+    )
+
+    @api.model
+    def get_default_stage(self):
+        return self.env['mgmtsystem.action.stage'].search([
+            ('is_starting', '=', True)
+        ]).id
```
dreispt, line 58: In Projects, the logic used is to pick the first non-folded Stage. Maybe we could also fall-back to that is no initial stage is set?

mgmtsystem_action/mgmtsystem_action.py
bwrsandman, line 51: I use lambda so `get_default_stage` can be overriden. AFAIK setting default to a member function directly does not allow it to be overriden the way that a call to `self.` does.

mgmtsystem_action/mgmtsystem_action.py
bwrsandman, line 58: Sure. Can you point me to the case in Projects?

dreispt, line 58: I could some changes myself: 
I would like to add Stages, and with that remove a lot of code that implemented a hardcoded process flow.
Also, do you actually use wkf workflows? If you let me I'm tempted to drop them also.

dreispt, line 58: I could some changes myself: 
I would like to add Stages, and with that remove a lot of code that implemented a hardcoded process flow.
Also, do you actually use wkf workflows? If you let me I'm tempted to drop them also.

bwrsandman, line 58: What is gained from dropping them? Workflows can be altered by the user which I find very useful.

dreispt, line 58: Changes in Workflow wiring are not reflected in the UI.
IMO Workflows are only useful for cross-Model processes (such as Sale -> Invoice -> Delivery).
For single-Model processes they are an overkill and can be replaced by a Kanban Board:

I would like to drop states and workflows, to replace them by Stages + Kanban Board.
History tracking and notifications are Stage will be there. Additional logic (mostly will be custom) is possible with Automated Actions.

bwrsandman, line 58: @max3903 do you agree?

max3903, line 58: I am sorry I don't. Workflows are useful to trigger server actions (like sending notifications).

dreispt, line 58: @max3903  Most common notifications are already triggered by mail_message tracking, and it's trivial to add additional messages or actions using Automated Actions. I have been using that extensively to: send reminder notifications, automatically route/assign issues to people, change kanban card colors to yellow or red or their priority, change block user actions on certain conditions.

I can share these implementation if you like. So I still don't see a use case where Workflows would be needed.

bealdav, line 58: I don't know your case but workflow seems overkill general performance of the erp. No ?
@dreispt :+1

max3903, line 58: @dreispt Ok, can you provide the required data and configuration to:
- send a notification to the responsible user when the action is open
- send a reminder to the responsible user 10 days before the action deadline

dreispt, line 58: @max3903 Agreed. The first one will be out of the box, through Stage tracking, the second will be a demo data example.

bwrsandman, line 58: Is this blocking the current PR or can we add @dreispt's ideas in a following PR?

bwrsandman:
@dreispt Most other porting PRs are dependent on this one, do you think your adjustments can be made after merging this?

max3903:
I think we added a name field on version 7 and I don't see it here.

Otherwise
:+1:

ghost:
Hi! @dreispt, @max3903 

May I include Spanish translation in this branch 8.0-mgmtsystem_nonconformity ?

max3903:
@gs-clearcorp Sure!

ghost:
PR here savoirfairelinux/management-system#18

dreispt:
@bwrsandman @max3903 Good enough today is better than perfect tomorrow, so yes, go ahead.
@max3903 I won't be able do to work on this, so maybe the idea can be put into the roadmap.

max3903:
@dreispt not sure what you want to add, but here it is #84 

:+1:

max3903:
:+1:

dreispt:
:+1:

PR #38, opened 2015-01-06
7.0 mgmtsystem nonconformity description

foutoucour:
Module to remove the spam description can produce in the view list and in the header of the form view


mgmtsystem_nonconformity_description/mgmtsystem_nonconformity.py
```diff
@@ -0,0 +1,29 @@
...
+##############################################################################
+
+from openerp.tools.translate import _
+from openerp.osv import fields, orm
+
+
+class mgmtsystem_nonconformity(orm.Model):
+    """ Add a field name to the nonconformity. """
+    _inherit = "mgmtsystem.nonconformity"
+    _rec_name = "ref"
```
bwrsandman: Why not add set ref as _rec_name directly in the `mgmtsystem_nonconformity` module?

dreispt:
Why a new module? Wouldn't it be best to modify `mgmtsystem_nonconformity` directly?

A question: would it be inappropriate to have a title and a description, as is usual in other models (Project Issues, for example)?

bwrsandman:
Same question as @dreispt

max3903:
:-1:

foutoucour:
Merging the update in the mgmsystem_nonconformity as suggested.

foutoucour:
Merging the update in the mgmsystem_nonconformity as suggested.

PR #39, opened 2015-01-06
7.0 mgmtsystem nonconformity field name

foutoucour:
Add the field name to nonconformities.
The name is also used to replace the description in the header of the form view.


max3903:
:-1: 

Just add the field in the current module

bwrsandman:
A whole module for a fix to nonconformity is probably overkill.

I should assume that these are fixes would benefit everyone using nonconformity.

It's probably better to update the existing module directly.

dreispt:
Isn't this solving the same thing as #38?
I suggest doing only one PR for these.

foutoucour:
Merging the update in the mgmsystem_nonconformity as suggested.

Issue #45, opened 2015-02-08
[7.0] mgmtsystem_probability


max3903:
This module allow you to manage scale of probability (or likelihood) used across different modules (health and safety, information security).
# Dependencies
- mgmtsystem
# Models

mgmtsystem.probability
- Name: char
- Value: integer
- Category: selection ('hazard', 'security')
# Menu

Menu item in Management System > Configuration > Probability
# Security
- MGMT User group can read
- MGMT Manager group can create, read, update and delete
# Data
- Maximum (every week), 4
- Strong (every month), 3
- Significant (every year), 2
- Minimum (every decade), 1

PR #47, opened 2015-02-10
[7.0] mgmtsystem_probability

llacroix:
https://github.com/OCA/management-system/issues/45

### DIFF
--- a/mgmtsystem_probability/README.rst
+++ b/mgmtsystem_probability/README.rst
@@ -1,0 +1,57 @@
+Management System Probability
+===========
+
+This module allow you to manage scale of probability (or likelihood)
+used across different modules (health and safety, information security).
+
+Installation
+============
+
+To install this module, you need to:
+
+ * Clone the project from github on your instance
+    * git clone https://github.com/OCA/management-system.git
+ * Start odoo with the project in the addons path
+
+Configuration
+=============
+
+No configuration needed
+
+Usage
+=====
+
+To use this module, you need to:
+
+ * go to Management System > Configuration > Probability
+
+For further information, please visit:
+
+ * https://github.com/OCA/management-system/issues
+
+Known issues / Roadmap
+======================
+
+ None
+
+Credits
+=======
+
+Contributors
+------------
+
+* Loïc Faure-Lacroix <loic.lacroix@savoirfairelinux.com>
+
+Maintainer
+----------
+
+.. image:: http://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: http://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose mission is to support the collaborative development of Odoo features and promote its widespread use.
+
+To contribute to this module, please visit http://odoo-community.org.
+
--- a/mgmtsystem_probability/README.rst
+++ b/mgmtsystem_probability/README.rst
@@ -1,5 +1,5 @@
 Management System Probability
-===========
+=============================
 
 This module allow you to manage scale of probability (or likelihood)
 used across different modules (health and safety, information security).
--- a/mgmtsystem_probability/__openerp__.py
+++ b/mgmtsystem_probability/__openerp__.py
@@ -1,0 +1,42 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    Copyright (C) 2015 - Present 
+#    Savoir-faire Linux (<http://www.savoirfairelinux.com>).
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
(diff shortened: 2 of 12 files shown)


max3903:
:+1:

mgmtsystem_probability/model/mgmtsystem_probability.py
```diff
@@ -0,0 +1,47 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+from openerp.tools.translate import _
+from osv import fields, orm
+
+_CATEGORIES = [
+    ("hazard", _("hazard")),
+    ("security", _("security")),
+]
```
dreispt: I'm uneasy with this being repeated in the modules.
It it's to be "used across different modules", maybe it could make sense to have a central table where the area specific modules would add the needed records.
Just an idea though... could be something to improved later.

llacroix: I agree with you, yet I'm not sure how such table would be named. Could be improved later as it's not yet clear if the model will stay as it is for long.

bwrsandman:
Is this still a WIP?

mgmtsystem_probability/model/mgmtsystem_probability.py
bwrsandman: `_("...")` wont work the way you expect it to here.

The function requires a `context` variable to be defined in the local frame.

Here is does not, since it is run when Odoo imports the module folder (therefore `__init__.py` and `mgmtsystem_probability.py`).

A way around this is to have a function reference in the `fields.selection(...)` declaration.

``` python
_columns = {
    "category": fields.selection(
        lambda self, *a, **kw: self._category_selection(*a, **kw),
        "Category", required=True,
    ),
}

def _category_selection(self, cr, uid, context=None):
    """Use a function so context is available"""
    return [
        ("hazard", _("hazard")),
        ("security", _("security")),
    ]
```

The use of lambda is to the function can be inherited and categories added.

See https://www.odoo.com/forum/help-1/question/how-to-extend-fields-selection-options-without-overwriting-them-21529#answer_21589 and my comments

bwrsandman:
Is this still a work in progress?

mgmtsystem_probability/model/mgmtsystem_probability.py
llacroix: > _("...") wont work the way you expect it to here.

I expected the parser that export translation to catch it. Openerp does the translation later. There is no need to call the translation method with the context at runtime because it will try to translate the translated string later probably.

But I kind of agree with the lambda part if you want to inherit it and update the list yet, the list could be static in the module anyway. Which mean that you don't have to recreate the list and translate the list multiple times.

llacroix: Hmm apparently if we use lambdas/methods it doesn't try to translate it twice.

llacroix:
Done not wip anymore

mgmtsystem_probability/__openerp__.py
```diff
@@ -0,0 +1,44 @@
...
+    "version": "1.0",
+    "author": "Savoir-faire Linux",
+    "website": "http://www.savoirfairelinux.com",
+    "license": "AGPL-3",
+    "category": "Management System",
+    "complexity": "normal",
+    "description": """\
+    This module allow you to manage scale of probability (or likelihood)
+    used across different modules (health and safety, information security).
+    """,
```
bwrsandman, line 89: Little inconsistent that you prevent a new line at the top with `\` of the description, yet add 4 spaces at each line.

``` python
"    This module allow you to manage scale of probability (or likelihood)\n    used across different modules (health and safety, information security).\n"
```

mgmtsystem_probability/model/mgmtsystem_probability.py
```diff
@@ -0,0 +1,68 @@
...
+
+    """
+    Define the Probability for management system.
+
+    Allow you to manage scale of probability (or likelihood)
+    used across different modules (health and safety, information security).
+    """
+
+    _name = "mgmtsystem.probability"
+    description = "Management System Probability"
```
bwrsandman: should be `_description`

mgmtsystem_probability/__openerp__.py
max3903, line 89: Please use the README.rst template from http://github.com/oca/maintainers-tools and remove the description from `__openerp__.py`.

bwrsandman, line 89: README.rst works on 7.0?

max3903, line 89: On Github and Odoo Apps, yes.

In a v7 instance, not sure.

llacroix, line 89: @bwrsandman I'm not sure what's really the matter here. Successive linebreaks and spaces are considered as one whitespace in html. So visually it wouldn't change anything to have it without the 4 spaces or the linebreaks.

llacroix, line 89: @max3903 No openerp7 doesn't support rst files. I can add the rst file and keep copy it to the description. At least it will work and be ready for v8

bwrsandman, line 89: Please do that

llacroix:
https://github.com/OCA/management-system/pull/56

yvaucher:
travis restarted

llacroix:
Yay :+1: runbot has the same problem with missing modules but travis is fine.

mgmtsystem_probability/README.rst
```diff
@@ -0,0 +1,57 @@
+Management System Probability
+===========
```
yvaucher: Some `=` are missing

llacroix:
Done

yvaucher:
:+1: once travis is green

llacroix:
green

yvaucher:
And merged thanks @llacroix

llacroix:
Thanks @yvaucher I just rebased #48 and did some similar changes to tests

PR #58, opened 2015-02-26
[FIX] flake8 error: E731 (removing lambdas)

llacroix:
https://travis-ci.org/OCA/management-system/jobs/52151402


llacroix:
Ah it was v8 not v7

dreispt:
I don't see why `E731` should be checked.
Using lambdas is part of developing Odoo.

lepistone:
See https://github.com/OCA/maintainer-quality-tools/issues/164 for a general discussion on the subject.

Issue #36, opened 2014-12-12
Print Verification List error


kfstar:
Hi,

I work for a company who is currently evaluating Open ERP and we have installed the Management system on Open ERP Version 7.0-20140415-231321 on Windows.

The management module is very good and this will be an important part of the software for us. 

I realise we do not have any support agreement however we have a problem with a part of the management system.

When we run the Print Verification List the following error is displayed, is there any easy fix for this error? : -

Many thanks and Regards,
Kevin

operator does not exist: integer = boolean
LINE 1: ...ge.id FROM "document_page" WHERE document_page.id IN (false)...
^
HINT: No operator matches the given name and argument type(s). You might need to add explicit type casts.

(<class 'psycopg2.ProgrammingError'>, ProgrammingError('operator does not exist: integer = boolean\nLINE 1: ...ge.id FROM "document_page" WHERE document_page.id IN (false)...\n ^\nHINT: No operator matches the given name and argument type(s). You might need to add explicit type casts.\n',), <traceback object at 0x0A2F2F58>)

max3903:
Workaround:
When you create your verification line, specify the procedure it is related to and you won't get this error message.

max3903:
@llacroix Can you fix it please ?

llacroix:
I'll work on it

dreispt:
Fix merged.

PR #63, opened 2015-04-01
[FIX][7.0] Return an undefined name for lines without document_page procedure

llacroix:
Fix the bug here:

https://github.com/OCA/management-system/issues/36


max3903:
:+1:

dreispt:
:+1:

Issue #71, opened 2015-05-07
mgmtsystem_information_security


max3903:
# Information Security Management System

This module adds a new management system and installs all the applications required to deploy and manage your Information Security Management System (ISMS), compliant with ISO 27001 standard.
# Installation

Before installing the module, you need to define the scope of your system and drive a security analysis. Method like EBIOS (http://en.wikipedia.org/wiki/EBIOS) can help you with the analysis, evaluation and action on risks relating to information systems.
# Configuration
- In the Configuration menu, define your Asset Categories and Threat Origins
- In the Manuals menu, define your Assets (Essentials and Underlying), Measures, Threat Scenarios, Security Events and Work Instructions
# Usage
- Plan Audits and Top Management Reviews
- Manage your Nonconformities and Actions
# Dependencies
- mgmtsystem_security_event
- information_security_manual
- mgmtsystem_audit
- mgmtsystem_review
# Data
- Add a new system 'Information Security' using the manual from information_security_manual

max3903:
- [ ] On the Action form view, the controls field should be displayed only when the system is of type 'information_system'. I tried with `attrs="{'invisible':[('system_id.type','!=', 'information_security')]}"` but using `system_id.type` does not work.

PR #48, opened 2015-02-10
[7.0] mgmtsystem_security_event

llacroix:
Related: https://github.com/OCA/management-system/issues/43
- [x] Model definition
- [x] Views for models
  - [x] Assets Category
  - [x] Essential Assets
  - [x] Underlying Assets
  - [x] Security Measure
  - [x] Threat Origin
  - [x] Threat Scenario
  - [x] Security Events - Scenario Lines
  - [x] Security Events - Measure Lines
  - [x] Security Events
- [x] Menus
  - [x] Security Events
  - [x] Configuration
- [x] Filter fields
  - [x] Security events
  - [x] Security Events - Scenario Lines
  - [x] Threat Scenario
  - [x] Security Measure
- [x] Name fields
  - [x] Scenario lines
  - [x] Measure lines
- [x] Tests
  - [x] security_assets_category.py
  - [x] security_assets_essential.py
  - [x] security_assets_underlying.py
  - [x] security_event.py
  - [x] security_event_measure.py
  - [x] security_event_scenario.py
  - [x] security_threat_origin.py
  - [x] security_threat_scenario.py
  - [x] security_measure.py
- [x] Translations [en, fr]

Dependencies to be merged:
- [x] mgmtsystem_severity https://github.com/OCA/management-system/pull/46
- [x] mgmtsystem_probability https://github.com/OCA/management-system/pull/47

PR #78, opened 2015-05-27
[ADD] ISMS module

max3903:
#71

Depends on
- [x] #48


max3903:
@dreispt The license key was missing in your modules. I took the liberty to set them to 'AGPL-3'. Let me know if you want to change them.

dreispt:
AGPL is the right thing.
Not many more options for < 9.0 AFAIK.

dreispt:
LGTM :+1:

PR #79, opened 2015-05-28
Add wkhtmltox to travis

ghost:


ghost:
@max3903 @bwrsandman I removed 3 unit tests in mgmtsystem_probability which fail. I don't agree with that kind of test because it might not even increase the coverage and it does not prove anything.

max3903:
:+1:

.travis.yml
```diff
@@ -30,6 +30,9 @@ install:
   - git clone https://github.com/OCA/maintainer-quality-tools.git ${HOME}/maintainer-quality-tools
   - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
   - travis_install_nightly
+  - apt-get install xfonts-75dpi
+  - wget http://sourceforge.net/projects/wkhtmltopdf/files/archive/0.12.2/wkhtmltox-0.12.2_linux-precise-amd64.deb
+  - dpkg -i wkhtmltox-0.12.2_linux-precise-amd64.deb
```
bwrsandman: You can't use `apt-get install` inside of a container travis. You need sudo.
Same for dpkg.

bwrsandman: Can you use one of these?
https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise#L6079
https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise#L6037

They can be added to the `addons: apt: packages:`

ghost: @bwrsandman I thought the ubuntu package had the version with unpatched qt. Is it still the case?

bwrsandman: I don't know... If that's the case, try downloading the binary locally and using an alias or something to avoid needing to use sudo.

ghost: @bwrsandman I am not sure what you mean or how to do that.

bwrsandman:
```
wget http://wkhtmltopdf.googlecode.com/files/wkhtmltopdf-0.11.0_rc1-static-amd64.tar.bz2
tar xj wkhtmltopdf-0.11.0_rc1-static-amd64.tar.bz2
```

then make the binary available to Odoo

ghost:
@bwrsandman Travis is green. I don't know why runbot is red.

pedrobaeza:
This are some warnings from deprecation type field in view and null constraint, so nothing about this PR.

I merge :+1:

PR #85, opened 2015-08-06
[UPD] Actions button doesn't change and log the status in the chatter

ehdem:
The button "Pending" doesn't work and the status is not log in the chatter


max3903:
Tested on runbot.
:+1:

dreispt:
LGTM :+1:

PR #88, opened 2015-08-23
[FIX] No such fields in mgmtsystem.nonconformity

max3903:
Fix Runbot warning


pedrobaeza:
There's still some warnings on runbot.

Can you please also use another technique for bringing another branch to reduce the commits to 1 for such trivial task (for example, `git cherry-pick`).

max3903:
Pushed

max3903:
Done in #89

PR #89, opened 2015-08-23
[FIX] No such field(s) in model mgmtsystem.nonconformity

max3903:
Fixes Runbot warnings.


max3903:
@pedrobaeza I don't know how to solve the other Runbot warnings:
- mgmtsystem_hazard already provides a default value
- mgmtsystem_nonconformity_deptm doesn't provide default values and I don't know which value @dreispt would like to set as default.

pedrobaeza:
Where are you seeing that warnings? These are the warnings I'm seeing:

```
2015-08-23 21:33:20,098 26929 WARNING 3113547-89-21ba30-all openerp.osv.orm: No such field(s) in model email.template: body_text.
2015-08-23 21:33:26,541 26929 WARNING 3113547-89-21ba30-all openerp.osv.orm: WARNING: unable to set column risk_computation_id of table res_company not null !
2015-08-23 21:33:30,550 26929 WARNING 3113547-89-21ba30-all openerp.osv.orm: WARNING: unable to set column department_id of table mgmtsystem_nonconformity not null !
2015-08-23 21:33:30,559 26929 WARNING 3113547-89-21ba30-all openerp.osv.orm: WARNING: unable to set column superior_user_id of table mgmtsystem_nonconformity not null !
```

max3903:
@pedrobaeza That's the ones I am talking about.
- First one is solved by this PR
- Second is my first point above: mgmtsystem_hazard already provides a default value
- Third and fourth ones are from @dreispt modules and there is no default values

pedrobaeza:
Let's see each of them:

> WARNING: unable to set column risk_computation_id of table res_company not null !

There's a method for getting a default, but you should use `self.pool['ir.model.data'].get_object_reference('mgmtsystem_hazard.risk_computation_a_times_b_times_c')[1]` for avoiding possible search troubles.

> WARNING: unable to set column department_id of table mgmtsystem_nonconformity not null !

mgmtsystem_nonconformity_deptm module should include 2 things:
- A default that searches for a department whose manager_id.user_id is the manager_user_id of the non conformity. If none, then I think the firtst department found can serve.
- Expand demo data in this module to complete this field on record with XML-ID `mgmtsystem_nonconformity.demo_nonconformity`
  
  > WARNING: unable to set column superior_user_id of table mgmtsystem_nonconformity not null !

If department, I would follow the same rule as the onchange, and if none found, the first found user. Demo data also needs to complete this field.

yvaucher:
:+1: Let's merge this one and then fix the other one next.

mgmtsystem_hazard/models/res_company.py
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    def _get_formula(self, cr, uid, context=None):
+        return self.pool['ir.model.data'].get_object_reference(
+            cr, uid,
+            'mgmtsystem_hazard', 'risk_computation_a_times_b_times_c',
+        )[1]
+
+    _defaults = {
+        'risk_computation_id': _get_formula
```
dreispt: I dislike names that don't reflect purpose. I would prefer something like `_default_risk_computation`, but I won't pick on that.

dreispt:
TravisCI is failing: 

```
ValueError: No such external ID currently defined in the system: mgmtsystem_hazard.risk_computation_a_times_b_times_c
```

max3903:
@dreispt Problem is that the data is not loaded yet. Any idea to work around this?

dreispt:
My best suggestion is: 
- make field not required in model
- make it required on the form view
- Add a function or post init hook to fill the values

max3903:
@naousse Can you implement @dreispt suggestions ? But instead of adding a function, let's just mention this configuration step in the README.rst.

PR #97, opened 2015-10-30
[9.0] information_security_manual

naousse:


information_security_manual/README.rst
```diff
@@ -0,0 +1,69 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License
+
+Information Security Management System Manual
+=============================================
+
+This module was written to provide an Information Security Manual Template.
```
elicoidal: This module provides an Information Security Manual Template.

max3903:
:+1:

information_security_manual/README.rst
```diff
@@ -0,0 +1,68 @@
...
+`here <https://github.com/OCA/management-system/issues/new?body=module:%20information_security_manual%0Aversion:%209.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+
+
+Credits
+=======
+
+Contributors
+------------
+
+* Odoo SA <info@odoo.com>
```
dreispt: Really?

information_security_manual/__openerp__.py
```diff
@@ -20,20 +20,18 @@
...
     "website": "http://www.savoirfairelinux.com",
     "license": "AGPL-3",
     "category": "Generic Modules/Others",
-    "description": """Information Security Manual Template
-    """,
     "depends": [
-        'mgmtsystem_manuals'
+        'mgmtsystem_manual'
     ],
     "init": [],
```
dreispt: deprecated, remove

dreispt:
1/ The module named should be prefixed with `mgmtsystem_`
2/ The module just adds a record on a model? Is that it?

max3903:
@dreispt 
1/ If we add the prefix, the name is tool long and truncated.
2/ The module adds the template of a ISMS manual based on ISO27001

dreispt:
You could use a shorter name; `mgmtsystem_infosec_manual`?
Unless I'm missing something, I see added only a documennt with a `<h1>Application Domain</h1>` title.

max3903:
@dreispt I think infosec is bad for SEO and yes, you missed the content of the category which hasn't changed:
https://github.com/naousse/management-system/blob/9.0-information_security_manual/information_security_manual/data/information_security_manual.xml

dreispt:
I still still we should have the module area prefix. I'm sure you can come up with a shorter module name.

max3903:
@dreispt Honestly, I don't think of any better one than yours...

@naousse Can you rename the module with Daniel's suggestion please?

dreispt:
@max3903 Since "infosec" is quite a specific keyword, what about `mgmtsystem_info_sec_manual`?

max3903:
@dreispt ping!

dreispt:
:+1:

Issue #104, opened 2015-11-04
ValueError: Invalid field 'fold' in leaf "<osv.ExtendedLeaf: ('fold', '=', False) on mgmtsystem_action_stage (ctx: )>"


ingfred:
# Impacted versions

First test on v8
# Steps to reproduce

Go to Reporting > Management Systems
# Results

```
File "/opt/odoo-8.0-20150129/openerp/osv/expression.py", line 812, in parse
    raise ValueError("Invalid field %r in leaf %r" % (left, str(leaf)))
ValueError: Invalid field 'fold' in leaf "<osv.ExtendedLeaf: ('fold', '=', False) on mgmtsystem_action_stage (ctx: )>"
```
# Expected results

The dashboard displays data without errors.

max3903:
@naousse can you look at this bug, write a test to catch it and fix it please?

PR #111, opened 2015-11-13
ValueError: Invalid field 'fold' in leaf "<osv.ExtendedLeaf: ('fold',…

naousse:
#104 issue solved


max3903:
I still have an error when opening the dashboard:
`Error: % undefined as prefix`

probably due to the "Open Claims" report as I don't have any message if I remove it.

PR #130, opened 2016-03-19
Update mgmtsystem_action.py

eugen-don:
Correct default Field description Value for Field company_id which should be Company and not System


max3903:
@eugen-don Thanks for the fix. Do you plan on signing the CLA?

eugen-don:
Hi Max yes, as soon as i get the time to read through the contract.

max3903:
@eugen-don too late, I applied your fix and gave you credits: 8bdd1464e74aa1abaf81c9848a3c40e543d8408a

PR #115, opened 2015-12-07
[9.0] mgmtsystem_action migration

naousse:
#105

PR #131, opened 2016-03-19
Update mgmtsystem_action.py

eugen-don:
Correct default Field description Value for Field company_id which should be Company and not System


max3903:
@eugen-don Thanks for the fix. Waiting for the CLA.

max3903:
Fixed in #115.

PR #140, opened 2016-06-09
9.0 mgmtsystem nonconformity

naousse:


mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -218,27 +227,71 @@ def create(self, vals):
 
     @api.multi
     def message_auto_subscribe(self, updated_fields, values=None):
```
naousse: @dreispt 
Please Daniel, i need your help here. I'm unable to register a nonconformity because of this method that doesn't work.

mgmtsystem_nonconformity/reports/mgmtsystem_nonconformity_report.py
```diff
@@ -0,0 +1,133 @@
...
+        'mgmtsystem_nonconformity_cause_rel',
+        'nonconformity_id',
+        'cause_id',
+        'Cause',
+    )
+    severity_id = fields.Many2one(
+        'mgmtsystem.nonconformity.severity',
+        'Severity', readonly=True,
+    )
+    origin_ids = fields.Many2many(
```
naousse: @dreispt My pivot view doesn't work when selecting a many2many field in the init method. Please i need help

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -218,27 +227,71 @@ def create(self, vals):
...
+        """
+        Add the responsible, manager and author to the OpenChatter follow list
+        """
         self.ensure_one()
         user_ids = [
             self.responsible_user_id.id,
             self.manager_user_id.id,
             self.author_user_id.id,
         ]
         self.message_subscribe_users(user_ids=user_ids, subtype_ids=None)
```
dreispt: You're passing as `user_ids` a list of users instead of a recordset. Chances are that this won't work.
I sugges to just make three calls to `message_subscribe_users()`

mgmtsystem_nonconformity/reports/mgmtsystem_nonconformity_report.py
dreispt: Why don't you create the Pivot view directly on top the `mgmtsystem.nonconformity`model?

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
naousse: @dreispt I have tried to do what you suggested. I still have the same error
That's the error i'm facing RuntimeError: maximum recursion depth exceeded while calling a Python object
I have also analyse the mail.thread object and i find that this code is correct from my point of view. Inside the mail.thread class, the user_ids is waiting a list of users id. If the list is None it builds a list with the id of the current users.

This is why only the current user is subscribe to the chatter when i deactivate this method. I hope the name of the throwing error will be helpfull. It seems that this error is linked to odoo 9

dreispt: OK, I suggest that you comment out that code and add a 
`# TODO` comment. We'll go ahead with the PR and in a later PR add the messaging logic.

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -120,7 +138,7 @@ def _state_name(self):
         track_visibility='onchange',
     )
     state_name = fields.Char(
```
dreispt: I don't think this fieds is needed. `State` alone is enough.

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -214,31 +279,75 @@ def create(self, vals):
...
+                vals.update(self.do_close())
+            if vals.get('state') == "cancel":
+                vals.update(self.do_cancel())
+            if vals.get('state') == "draft":
+                vals.update(self.case_reset())
+
+        result = super(MgmtsystemNonconformity, self).write(vals)
+        return result
+
+    def checkifTheProcessIsClosedOrCancelled(self):
```
dreispt: We shouldn't really use CamelCase.

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -19,7 +19,7 @@
 #
 ##############################################################################
 
-from openerp import models, api, fields, exceptions, _
+from openerp import models, api, fields, netsvc, exceptions, _
```
dreispt: Why is netsvc needed?

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -263,7 +260,7 @@ def _elapsed_days(self, dt1_text, dt2_text):
             res = (dt2 - dt1).days
         return res
 
-    def _compute_state_name(self):
+    def _state_name(self):
```
dreispt: I can't see why we need to compute the state name - it is a selection list!

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -289,8 +287,11 @@ def message_auto_subscribe(self, updated_fields, values=None):
         """
         self.ensure_one()
 
-        # user_ids = [self.responsible_user_id._uid, self.manager_user_id._uid,
-        #            self.author_user_id._uid]
+        # user_ids = [
+        #    self.responsible_user_id._uid,
+        #    self.manager_user_id._uid,
+        #    self.author_user_id._uid,
+        # ]
```
dreispt: remove unused code

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -318,15 +319,15 @@ def write(self, vals):
         result = super(MgmtsystemNonconformity, self).write(vals)
         return result
 
-    def checkifTheProcessIsClosedOrCancelled(self):
+    def check_if_the_process_is_closed_or_cancelled(self):
```
dreispt: you can make these shorter: `check_closed_or_cancelled`

max3903:
@naousse You can rebase this branch now.

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
naousse: I left this code because because, i thouht it can be useful when we will find how  to automatically register those users as followers.

max3903:
@naousse 
- [x] The create button has disappeared when you click on nonconformity in the menu
- [x] Drag and drop in the kanban view is not working
- [x] Merge https://github.com/naousse/management-system/pull/7 to improve README file
- [x] The "Close" button has disappeared. I had to click on "Closed" in the workflow steps to close the NC
- [x] Check the form view on branch 8.0, I think the action plan comments and effectiveness review fields have been improved

dreispt:
@naousse Let us know when it's ready to review.

naousse:
@max3903 
I choose to remove the close button, because i think the workflow steps is enough. But i will add it if you find it useful. The form view in 8.0 branch is no different from what i have. That's why there is no change

@dreispt 
The code is ready for review.
I just have 4 problems. 
- I didn't find out how to add automatically follower in chatter
- I didn't find how to add many2many column in pivot view
- I use the code you suggested to calculate measure in model, but today i realise that it is not working
- The not tested code is because i have to switch from admin user to demo user to test that code and i don't know how to do it.

Please that's where i need your help.

max3903:
@naousse Please double check the view, I think you are missing theses 2 commits:
https://github.com/OCA/management-system/commit/285f30d0e4a06b04da0ad995a63122e99f972bce
https://github.com/OCA/management-system/commit/82c2e1d67ab970540c2e198dbe36c5f3cdbcbbe6

dreispt:
>    I didn't find out how to add automatically follower in chatter

The `user_id` is automatically added.
But don't worry too much - add the Task to the  README "Roadmap" and we can implement that in a later PR.

>    I didn't find how to add many2many column in pivot view

AFAIK you can't. Only m2o.

>    I use the code you suggested to calculate measure in model, but today i realise that it is not working

Only step-by-step debugging can figure that out.

>    The not tested code is because i have to switch from admin user to demo user to test that code and i don't know how to do it.

Here is one way to do it: https://github.com/dreispt/todo_app/blob/9.0/todo_app/tests/test_todo.py#L13

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -204,6 +241,47 @@ def _state_name(self):
         domain="[('nonconformity_id', '=', id)]",
     )
 
+    def _compute_age(self):
```
dreispt: missing `@api.depends('create_date')`

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -73,28 +83,58 @@ def _state_name(self):
...
+    state_id = fields.Many2one(
+        'mgmtsystem.nonconformity.state',
+        'State',
+        default=_default_state)
+
+    # Compute data
+    number_of_nonconformities = fields.Integer(
+        '# of nonconformities', readonly=True, default=1)
+    age = fields.Integer('Age', readonly=True,
+                         compute='_compute_age', store=True)
```
dreispt: This can't have `store=True`, since the computation depends on current date.
As is will store zero all the time.

max3903:
ping @naousse

max3903:
@naousse Can you update the view with the 2 commits I provided above?

dreispt:
@naousse @max3903 I'm afraid this train is going late...

max3903:
@naousse I don't know how many speeds you got on your train, but right now is the time to get to the latest ;)

max3903:
:+1:

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -73,28 +83,58 @@ def _state_name(self):
         readonly=True,
         default="NEW"
     )
-    date = fields.Date(
-        'Date',
+
+    state_id = fields.Many2one(
+        'mgmtsystem.nonconformity.state',
+        'State',
+        default=_default_state)
```
dreispt: - [ ] Why did we replace a Selection list with a Many2one?

For consistency with official modules, you either have a `stage` field with a selection list, on you have Many2one  `stage_id`  (thus using the term Stage instead of State).

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -29,40 +29,50 @@
...
+
+    _STATES = [
+        ('draft', _('Draft')),
+        ('analysis', _('Analysis')),
+        ('pending', _('Pending Approval')),
+        ('open', _('In Progress')),
+        ('done', _('Closed')),
+        ('cancel', _('Cancelled')),
+    ]
+    _STATES_DICT = dict(_STATES)
```
dreispt: Remove States if you choose to use Stages.

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -212,33 +290,97 @@ def verbose_name(self):
...
+                    "mgmtsystem_nonconformity.state_done").id:
+                vals.update(self.do_close())
+            if vals.get('state') == "cancel" or vals.get(
+                    'state_id') == self.env.ref(
+                    "mgmtsystem_nonconformity.state_cancel").id:
+                vals.update(self.do_cancel())
+            if vals.get('state') == "draft" or vals.get(
+                    'state_id') == self.env.ref(
+                    "mgmtsystem_nonconformity.state_draft").id:
+                vals.update(self.case_reset())
```
dreispt: Looking at this, you should keep using States and not move to Stages.
There is a solution to have the best of both worlds, but let's leave that to a future iteration.

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
dreispt: States are used in write logic, so it is easier if you stick to them.
IMO you should keep states as a selection list, as you had before.

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
dreispt: OK, add a To-do for this to the roadmap section of the README, if you don't mind.

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
naousse: Please @dreispt. I decided to add stage because of Kanban View. With State it was impossible to have drag and drop in kanban view. I search how to solve it without success. That is why i decided to go that way. If i remove stage, i don't know how to add drag and drop in kanban view with a simple field.
If it is Ok i can remove state and just work with stage

naousse: @dreispt i apply your advice to change the current user in order to test the part of code that need a non administrator user without success.  Thta is why the coverage doesn't reach 100%

dreispt: I suggest to stick to states for now.
We can add Stages in a later PR.
In fact, there is a [WIP module](https://github.com/OCA/server-tools/pull/368) to automatically add that to an exiting model - we can leverage that when ready.

dreispt: If you still prefer to use Stages (and that is the proper solution for Kanbans) you will still need to keep a computed `stage` field, so that the logic works properly.

This means that you have a `stage_id` on conformities, and a Stages tables.
And then you must map each Stage to a State, and have a `related` field on Nonconformities that automatically computes it's `state`  from the document's `stage_id`.
I implemented this for Projects, here: https://github.com/OCA/project/tree/8.0/project_stage_state

naousse: As you suggestted, It is Ok for me too to keep it  for now

max3903:
@dreispt What's missing to merge? Can you create issues for the stuff that can be postponed to a later PR?

dreispt:
@max3903 AFAICT States are broken right now, the `state` field used in the logic is not present.
@naousse Please make `state` a Selection field again, and IMO state_id should be removed. 
In a later PR we should introduce `stage_id`.

naousse:
@dreispt 
State is still there and it is a selection fields. I just added state_id for us to have  drag and drop in kanban view
state_id doesn't replace state. the both are there. There is no place where state_id replace state. State_id is just a special field for kanban view drag and drop

max3903:
@dreispt If @naousse renames `state_id` by `stage_id`, would that work for you?

dreispt:
@max3903 @naousse I'm working on it.

naousse:
@dreispt 
I have made some modification and i'm pushing it .
Mabe you should check if it what you expected
We now have state and stage_id

dreispt:
Thnks. As I mentioned, I was working on it - I just created https://github.com/naousse/management-system/pull/9 with the WIP. I should now fix conflicts with your changes and I still need to do some changes to the tests.

@max3903 I removed all the workflow logic - it was added by me and now I think it is imposing a particular workflow: it should rather be as workflow agnostic as possible, to then let people add the workflow best fit for their use cases. 
Doing so also simplifies a lot the code in the module.

dreispt:
@naousse 
Hi, https://github.com/naousse/management-system/pull/9 is ready to merge here:
- fully implemented `state` derived from Stages
- removed the probably too specific workflow logic, so that we can freely implement the business logic we need on top of this.
- simplified to tests, accordingly
- added the `kanban_state`. Is can be used to indicate the need to approval ("blocked") / approved sub-states. This can be leveraged by Automated Action, for example.

@max3903 
I still dislike the Report provided , and think it should instead be a Pivot view on non-conformities.
But I won't pick on that if you insist keeping it.

max3903:
@naousse 
- [x] The report view is not working for other measures than count. I get an error message

```
ProgrammingError: column mgmtsystem_nonconformity_report.number_of_days_to_analyse does not exist
LINE 2: ...onformities") AS "number_of_nonconformities" ,sum("mgmtsyste...
```
- [x] I was able to close a NC by drag and drop without the effectiveness review. I should get an error message and be asked to enter it.

@dreispt I am not sure to understand your comment: the report is a pivot view on nonconformities.

dreispt:
@naousse I removed the workflow logic, I can handle the needed validations and others topics, except the report related one.

@max3903 The report is based on a different model, using a database view, with only a subset of the NC data. IMO it should be a Pivot view directly on the NV model/table.

max3903:
@naousse What is the reason for this new model? Why can't we base the pivot view directly on the NC object?

naousse:
@max3903, @dreispt 
The new model is needed because i want to calculate the age  of a nonconformity. When relying on NC object, the pivot view doesn't see that field because it is a compute field. For pivot to see the field we have to add the parameter store=True. But because the age depends on a field that value doesn't change, it never calculate the age of the NC. That's why i decided to build a view that will calculate the age each time, the pivot view is called

dreispt:
@naousse That's a good explanation.
@max3903 Is the absolute age relevant? AFAICS the cycle time for a NC can vary wildly, depending on it's specifics...

max3903:
@dreispt I think it is. External auditors want you to fix the oldest NC first. They don't like NC that remains open in the system for years...

dreispt:
@max3903 There are other ways to achieve that, and that particular computation won't work since it can't be stored.
We have some options:
a) Instead of using a field computation, have a daily cron job update the age of the open NCs
b) Use a "Days since last update" computed field instead. It allows to track idle NCs and can be made a stored field.

max3903:
@dreispt Option b

dreispt:
My changes are ready on https://github.com/naousse/management-system/pull/10

naousse:
@dreispt 
Please I try to caculate the days since last update, but i'm not finding how to do it. What are the field involved. And on which field will depends the compute method?

dreispt:
@max3903 I didn't get this one:

> When the NC is in progress state/stage, the actions of the plan should be opened.

dreispt:
@naousse https://github.com/naousse/management-system/pull/11 implements it and a few other improvements/fixes - check the commit messages

max3903:
- [x] I was able to move the NC to the In Progress state/stage without the plan review. I should get an error message and be asked to enter it.
- [x] In the chatter of a nonconformity, the state/stage changes appears twice in each message
- [x] When the NC is in progress state/stage, the actions of the plan should be opened.
- [x] The menu item for report is not consistent with the others: `Nonconformity` --> `Nonconformities`

See http://3151194-140-e585e9.runbot2.odoo-community.org/web#id=2&view_type=form&model=mgmtsystem.nonconformity&menu_id=131&action=120

dreispt:
> When the NC is in progress state/stage, the actions of the plan should be opened.

I don't understand this.

max3903:
@dreispt Actions are in draft until the action plan is reviewed. Once done, actions are opened and assignees can start working on them.

dreispt:
There you go: https://github.com/naousse/management-system/pull/12 should fix those issues.
During tests, I figured that the "Pending Approval" stage wasn't intuitive, so I renamed it to "Action Plan".

max3903:
- [x] ￼ I was able to close the NC with one action opened. I should get an error message and be asked to close every actions first (the immediate action and the ones of the action plan).

Otherwise :+1:

dreispt:
There you go https://github.com/naousse/management-system/pull/13

dreispt:
Coverage back to 100% 😺

max3903:
:+1: Good work guys!

dreispt:
🏆

PR #148, opened 2016-07-22
[8.0] Remove trivial tests in mgmtsystem_nonconformity

ghost:
These unit test do not test anything. Raising an exception when a required attribute is missing must never be tested because this is already tested in the framework. The result is that an error appears in the logs when testing the module.


max3903:
:+1:

dreispt:
:+1:
@naousse Please include change in #140. You can cherry-pick the commit.

dreispt:
PS: I'll work on these tests for 9.0, so I'll handle this.

PR #149, opened 2016-07-27
[8.0] Fix mgmtsystem_action case_open

ghost:
This method could lead to "Expected singleton" error.
Nothing ensures that there is only one state that is not the start nor the end.

### DIFF
--- a/mgmtsystem_action/mgmtsystem_action.py
+++ b/mgmtsystem_action/mgmtsystem_action.py
@@ -85,7 +85,7 @@ class mgmtsystem_action(models.Model):
             values['stage_id'] = stages.search([
                 ['is_ending', '=', False],
                 ['is_starting', '=', False]
-            ]).id
+            ], limit=1, order='sequence').id
 
             case.write(values)
 
--- a/mgmtsystem_action/__openerp__.py
+++ b/mgmtsystem_action/__openerp__.py
@@ -20,7 +20,7 @@
 ##############################################################################
 {
     "name": "Management System - Action",
-    "version": "8.0.1.2.1",
+    "version": "8.0.1.2.2",
     "author": "Savoir-faire Linux,Odoo Community Association (OCA)",
     "website": "http://www.savoirfairelinux.com",
     "license": "AGPL-3",


mgmtsystem_action/mgmtsystem_action.py
```diff
@@ -80,13 +80,8 @@ def case_open(self):
...
             values = {'active': True}
-
-            stages = self.env['mgmtsystem.action.stage']
-            values['stage_id'] = stages.search([
-                ['is_ending', '=', False],
-                ['is_starting', '=', False]
-            ]).id
-
+            values['stage_id'] = self.env.ref(
+                'mgmtsystem_action.stage_open').id
```
dreispt: It's not equivalent code, since you could use custom Stages, with no XML Id.
I suggest instead to add `limit=1` to the original `search()`.

ghost: @dreispt I don't agree. If you add limit, you get a non-deterministic behavior.

And the name of the method is "case_open". So the logic state of the action is "open".

dreispt: The Stages list is editable.
I can scrap all the default stages and create new ones.
In that case, your fix wouldn't work, since there is no Stage with that XML ID.

But the flaw is in the design: "Open" and "Close" buttons worked well on closed lists of States.
When we evolve to configurable Stages, they are no longer useful, since we can perfectly have several "open" Stages in our Kanban.
The solution would be to drop those buttons, and have the user move the Action the the exact Stage they want.

ghost: @dreispt I agree, this is a flaw of design. I will fix it and update the pull request.

dreispt: It might be worth taking a look at what was made on the port to 9.0.
Maybe backporting those changes is a sensible thing to do.

ghost: @dreispt I don't have a lot of time to spend on this issue and backporting could affect other people who use the module. What about adding limit=1 and order='sequence' ?

ghost:
@dreispt how about this ?

dreispt:
:+1: for the quickfix

ghost:
@dreispt thanks for the review

max3903:
@dufresnedavid can you bump the version number?

:+1:

ghost:
@max3903 done

PR #143, opened 2016-06-24
mgmtsystem_audit progress

naousse:

PR #152, opened 2016-08-01
[MIG] mgmtsystem_audit

max3903:
Fixes #143 conflicts.


max3903:
@naousse Few comments:
- [x] Separate the verification line class in its own file. One class per file.
- [x] If we delete the audit, we want to remove the verification lines, but keep the procedures. Not sure the ondelete are correct.
- [x] `Audits` menu items should be below `Nonconformities`

In the Verification List report ([See attached)](https://github.com/OCA/management-system/files/391619/Verification.list.report.2.pdf):
- [x] `Confidentially` --> `Confidentiality`
- [x] Date value is not aligned with the reference and confidentiality ones
- [x] Columns content is shifted to the right. Comments column should be empty. 

In the demo data:
- [x] Convert the `Strong points` and `To improve points` values into HTML

max3903:
:+1:

dreispt:
:+1:

PR #166, opened 2017-03-14
MIG mgmtsystem_system to V. 10.0

eugen-don:
MIG mgmtsystem_system to V. 10.0

PR #168, opened 2017-03-14
Mig mgmtsystem action to v. 10.0

eugen-don:
Mig mgmtsystem action to v. 10.0
depends: 
- [x] https://github.com/OCA/management-system/pull/166 mgmtsystem

### DIFF
--- a/mgmtsystem_action/models/mgmtsystem_action.py
+++ b/mgmtsystem_action/models/mgmtsystem_action.py
@@ -1,10 +1,10 @@
 # -*- coding: utf-8 -*-
 
-from openerp import fields, models, api, exceptions, _
+from odoo import fields, models, api, exceptions, _
 from datetime import datetime, timedelta
 
 
-class MgmtSystemAction(models.Model):
+class MgmtsystemAction(models.Model):
     """Model class that manage action."""
 
     _name = "mgmtsystem.action"
@@ -85,24 +85,14 @@ class MgmtSystemAction(models.Model):
     stage_id = fields.Many2one(
         'mgmtsystem.action.stage',
         'Stage',
-        default=_default_stage)
+        track_visibility='onchange', index=True,
+        copy=False,
+        default=_default_stage, group_expand='_stage_groups')
 
     @api.model
-    def _stage_groups(self, present_ids, domain, **kwargs):
-        """This method is used by Kanban view to show empty stages."""
-        # perform search
-        # We search here only stage ids
-        stage_ids = self.env['mgmtsystem.action.stage']._search([])
-        # We search here stages objects
-        result = self.env['mgmtsystem.action.stage'].search([]).name_get()
-        # restore order of the search
-        result.sort(lambda x, y: cmp(
-            stage_ids.index(x[0]), stage_ids.index(y[0])))
-        return result, None
-
-    _group_by_full = {
-        'stage_id': _stage_groups
-    }
+    def _stage_groups(self, stages, domain, order):
+        stage_ids = self.env['mgmtsystem.action.stage'].search([])
+        return stage_ids
 
     @api.model
     def _get_stage_new(self):
@@ -136,7 +126,7 @@ class MgmtSystemAction(models.Model):
         """Creation of Action."""
         Sequence = self.env['ir.sequence']
         vals['reference'] = Sequence.next_by_code('mgmtsystem.action')
-        action = super(MgmtSystemAction, self).create(vals)
+        action = super(MgmtsystemAction, self).create(vals)
         self.send_mail_for_action(action)
         return action
 
@@ -184,14 +174,13 @@ class MgmtSystemAction(models.Model):
                     body=' %s ' % (_('Action cancelled on ') +
                                    fields.Datetime.now())
                 )
-        return super(MgmtSystemAction, self).write(vals)
+        return super(MgmtsystemAction, self).write(vals)
 
+    @api.model
     def send_mail_for_action(self, action, force_send=True):
-        """Set a document state as draft and notified the reviewers."""
         template = self.env.ref(
             'mgmtsystem_action.email_template_new_action_reminder')
-        for action in self:
-            template.send_mail(action.id, force_send=force_send)
+        template.send_mail(action.id, force_send=force_send)
         return True
 
     def get_action_url(self):
@@ -219,5 +208,5 @@ class MgmtSystemAction(models.Model):
         template = self.env.ref(
             'mgmtsystem_action.action_email_template_reminder_action')
         for action in action_ids:
-            template.send_mail(action.id, force_send=True)
+            template.send_mail(action.id)
         return True
--- a/mgmtsystem_action/models/mgmtsystem_action_stage.py
+++ b/mgmtsystem_action/models/mgmtsystem_action_stage.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-from openerp import models, fields
+from odoo import models, fields
 
 
 class MgmtsystemActionStage(models.Model):
--- a/mgmtsystem_action/views/mgmtsystem_action_stage.xml
+++ b/mgmtsystem_action/views/mgmtsystem_action_stage.xml
@@ -1,5 +1,4 @@
 <odoo>
-  <data>
 
     <record model="ir.ui.view" id="mgmtsystem_action_stage_tree_view">
       <field name="name">Management System action Stage Tree</field>
@@ -22,7 +21,4 @@
       <field name="context">{}</field>
     </record>
 
-
-
-  </data>
 </odoo>
--- a/mgmtsystem_action/README.rst
+++ b/mgmtsystem_action/README.rst
@@ -63,7 +63,7 @@ In case of trouble, please check there if your issue has already been reported.
 If you spotted it first, help us smashing it by providing a detailed and welcomed feedback `here <https://github.com/OCA/
 Management-system/issues/new?body=module:%20
 mgmtsystem_system%0Aversion:%20
-9.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+10.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
 
 
 Credits
@@ -74,6 +74,7 @@ Contributors
 
 * Savoir-faire Linux <support@savoirfairelinux.com>
 * Gervais Naoussi <gervaisnaoussi@gmail.com>
+* Eugen Don <eugen.don@don-systems.de>
 
 Maintainer
 ----------
@@ -92,6 +93,7 @@ To contribute to this module, please visit http://odoo-community.org.
 
 Changelog
 ---------
+
 
 v9.0.1.0.0
 
--- a/mgmtsystem_action/__init__.py
+++ b/mgmtsystem_action/__init__.py
@@ -1,4 +1,6 @@
 # -*- coding: utf-8 -*-
 
-from . import models
-from . import reports
+from . import (
+    models,
+    reports,
+)
--- a/mgmtsystem_action/__manifest__.py
+++ b/mgmtsystem_action/__manifest__.py
@@ -1,26 +1,10 @@
 # -*- coding: utf-8 -*-
-##############################################################################
-#
-#    OpenERP, Open Source Management Solution
-#    Copyright (C) 2010 Savoir-faire Linux (<http://www.savoirfairelinux.com>).
-#
-#    This program is free software: you can redistribute it and/or modify
-#    it under the terms of the GNU Affero General Public License as
-#    published by the Free Software Foundation, either version 3 of the
-#    License, or (at your option) any later version.
-#
-#    This program is distributed in the hope that it will be useful,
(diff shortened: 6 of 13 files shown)


sbidoul:
I think this one is the next in the dependency graph.

I relaunched travis which shows a lint error and a couple of test failures.

eugen-don:
@sbidoul thanks

eugen-don:
@sbidoul 
the joblogs not showing since the relaunch?

edit:
Fixed linting errors in reports and tests ... rebasing on OCA/10.0

eugen-don:
rebased on OCA/10.0

mgmtsystem_action/models/mgmtsystem_action.py
```diff
@@ -136,15 +136,25 @@ def create(self, vals):
         """Creation of Action."""
         Sequence = self.env['ir.sequence']
         vals['reference'] = Sequence.next_by_code('mgmtsystem.action')
-        action = super(MgmtSystemAction, self).create(vals)
-        self.send_mail_for_action(action)
```
grindtildeath: Why not send the email ?

mgmtsystem_action/models/mgmtsystem_action.py
```diff
@@ -136,15 +136,25 @@ def create(self, vals):
...
+            res['stage_id'] = (
+                self.stage_id.mail_template_id,
+                {'composition_mode': 'mass_mail'})
+        return res
+
+    @api.multi
     def write(self, vals):
         """Update user data."""
         if vals.get('stage_id'):
-            stage_new = self._get_stage_new()
```
grindtildeath: Why change method names ?

mgmtsystem_action/models/mgmtsystem_action.py
```diff
@@ -19,11 +33,14 @@ def _default_owner(self):
...
         return self.env.user
 
+    @api.model
     def _default_stage(self):
         """Return the default stage."""
-        return self.env['mgmtsystem.action.stage'].search(
-            [('is_starting', '=', True)],
-            limit=1)
+        return (
+            self.env.ref('mgmtsystem_action.stage_draft', False) or
```
grindtildeath: Why change the default behaviour ? I suppose we should be able not to use stage_draft as default (is_starting=True).

mgmtsystem_action/models/mgmtsystem_action.py
```diff
@@ -1,15 +1,29 @@
...
+# Copyright (C) 2010 Savoir-faire Linux (<http://www.savoirfairelinux.com>).
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from openerp import fields, models, api, exceptions, _
+from odoo import models, api, fields, exceptions, _
 from datetime import datetime, timedelta
 
 
-class MgmtSystemAction(models.Model):
+class MgmtsystemAction(models.Model):
```
grindtildeath, line 7: Please don't change class names for compatibility.

mgmtsystem_action/models/mgmtsystem_action.py
```diff
@@ -1,15 +1,29 @@
...
     """Model class that manage action."""
 
     _name = "mgmtsystem.action"
     _description = "Action"
+    _rec_name = "name"
     _inherit = ['mail.thread', 'ir.needaction_mixin']
+    _order = "date_deadline desc"
+    _track = {
```
grindtildeath: What is this supposed to do ?

mgmtsystem_action/models/mgmtsystem_action_stage.py
```diff
@@ -1,21 +1,59 @@
...
+
+    name = fields.Char(string='Stage Name', required=True, translate=True)
+    case_default = fields.Boolean('Common to All Teams')
+    is_starting = fields.Boolean(
+        string='Is starting Stage',
+        help="select stis checkbox if this is the default stage \n"
+             "for new nonconformities")
     is_ending = fields.Boolean('Ending stage')
+    sequence = fields.Integer(
+        'Sequence', default=1, help="Used to order stages. Lower is better.")
```
grindtildeath: Please keep default behaviour (default=100)

mgmtsystem_action/models/mgmtsystem_action_stage.py
```diff
@@ -1,21 +1,59 @@
...
-        It will not assign this stage to existing teams.")
-    is_starting = fields.Boolean('Starting stage')
+    def _get_mail_template_id_domain(self):
+        return [('model', '=', 'mgmtsystem.action')]
+
+    name = fields.Char(string='Stage Name', required=True, translate=True)
+    case_default = fields.Boolean('Common to All Teams')
+    is_starting = fields.Boolean(
+        string='Is starting Stage',
+        help="select stis checkbox if this is the default stage \n"
```
grindtildeath: "Select this checkbox if this is default stage." (it might for for something else than nonconformities)

mgmtsystem_action/views/mgmtsystem_action_stage.xml
```diff
@@ -1,28 +1,95 @@
...
-        </tree>
-      </field>
-    </record>
+        <record id="mgmtsystem_action_stage_tree_view" model="ir.ui.view">
+            <field name="name">mgmtsystem.action.stage.tree</field>
+            <field name="model">mgmtsystem.action.stage</field>
+            <field name="arch" type="xml">
+                <tree string="Action Stage">
+                    <field name="name"/>
+                    <field name="sequence" widget="handle" groups="base.group_no_one"/>
```
grindtildeath: Why define this group here ?

grindtildeath (commented):
Only code review. 
Just didn't review all the new features as I don't know what it's supposed to do, and I'm not sure a migration PR is the best place for it.

sbidoul:
@eugen-don thanks for your work around here. 

To facilitate review and accelerate the migration, perhaps could you do a first PR for the migration, and then a second one with the new features?

mgmtsystem_action/models/mgmtsystem_action.py
eugen-don: @grindtildeath 
The prior version of the method returned the xmlid of the stage, stage_draft, meaning the stage with id stage_draft had to be present at all times in the database for the create method to work, now if a manager user deletes the stage with the id stage_draft it can no longer be retrieved by the method and create method for new records fails. So this new method checks for there being a stage with id stage_draft and if this stage is not present it returns the stage that has the bool field as being the starting stage.

mgmtsystem_action/models/mgmtsystem_action.py
eugen-don: @grindtildeath 
Did you try it out? The modules names are seperated by a dot . 

So in class MgmtsystemAction
field _name is written as name = "mgmtsystem.action"
the casecamel has to bee accordingly or it throwse an error

eugen-don:
@sbidoul @grindtildeath 

I see the new features cause some confusion on the review.

ILl split the code this week in 2 PRs one just the migration and an aditional one for the new features..

mgmtsystem_action/models/mgmtsystem_action.py
eugen-don: @grindtildeath 
That was a sunday miskate, thanks for the spot my friend :D

mgmtsystem_action/views/mgmtsystem_action_stage.xml
eugen-don: @grindtildeath 
thats also not needed, since ill be splitting migration code and new features, this will be left out as well. 

Thanks again for the good review...
Also thanks to @sbidoul for spotting the linting errors and the test failures.
As for the failures im still working on it.

mgmtsystem_action/models/mgmtsystem_action.py
eugen-don: Correction: This was not an mistake, this is removed because of the new feature introduced in project module in V10.0, which I adapted in this module.

Its this method: _track_template(self, tracking):
checks if stage_id has been changed and sends the mail template that is associated to the stage.

Big advantage in comparison to the old method is that you can choose which stages send notifications for actions. In the old method the notifications were allways send regardles if you wanted it or not.

mgmtsystem_action/models/mgmtsystem_action.py
eugen-don: Hi @grindtildeath 
I changed the methods name because stage_new() implies creating a new stage, the name does not imply on the methods intention to retrive the default stage from the database.

So if possible i would like to stay consistent to the odoo core modules with stage_default or something like get_default_stage_id so that the methods name at least contains the word default in it.
I chose _default_stage since this was the shortest method name fullfillig the requirement of not beein missleading.

/opt/odoo/10/srv/addons/crm/tests/test_lead2opportunity.py:            'stage_id': default_stage_id

sbidoul:
> ILl split the code this week in 2 PRs one just the migration and an aditional one for the new features..

@eugen-don thanks! Let me know when it's done, it's should be mergeable quickly so we can move forward with the rest while the feature improvement is fine-tuned.

grindtildeath:
@eugen-don 
All right let me know when PR is split and will review one more time. Huge thanks for your work on this module by the way.

max3903 (changes requested):
When the action is closed, all the fields should be readonly.

Otherwise :+1:

eugen-don:
Hi @max3903

it allready was fixed but i removed it since some people told me to put improvements in a different PR.

IMO its an improvement since in v 9.0 you can edit all fields in closed actions.

actions_v9


Should i first make a fix for v 9.0 and do a PR there? or else we have new code here that cant be compared against v 9.0 when reviewing...

sbidoul:
@eugen-don IMO keep improvements in other PRs so we can expedite migration and start collaborating on improvements.

mgmtsystem_action/models/mgmtsystem_action.py
```diff
@@ -184,14 +174,14 @@ def write(self, vals):
...
 
+    @api.multi
     def send_mail_for_action(self, action, force_send=True):
-        """Set a document state as draft and notified the reviewers."""
         template = self.env.ref(
             'mgmtsystem_action.email_template_new_action_reminder')
-        for action in self:
-            template.send_mail(action.id, force_send=force_send)
+        self.env['mail.template'].browse(template.id).send_mail(
+            action.id, force_send=force_send)
```
sbidoul: @eugen-don why do you rmove the `for action` loop?

eugen-don: @sbidoul Hi
The mails are send on record creation, there is no need for a looping.
I actually wanted to preserve the looping since i wasnt sure about its purpose... 

So i did some extensive testin just to figure out that the looping prevented the mails from beeing send on record creation

sbidoul: Ah I see it receives the action as argument. Then it must be `@api.model` since this method does not act on a recordset.

max3903 (approved):

mgmtsystem_action/models/mgmtsystem_action.py
eugen-don: you think we should use model here instead of multi?

sbidoul: yes, it's mechanical: if you have @api.multi, you must have either ensure_one or iterate over the recordset, in this case you don't so @api.model correctly expresses this is a method that does not act on a recordset of actions

eugen-don: Thanks... thats a good explanation... ill do a quick test and change this

mgmtsystem_action/models/mgmtsystem_action.py
```diff
@@ -184,14 +174,14 @@ def write(self, vals):
...
+        return super(MgmtsystemAction, self).write(vals)
 
+    @api.multi
     def send_mail_for_action(self, action, force_send=True):
-        """Set a document state as draft and notified the reviewers."""
         template = self.env.ref(
             'mgmtsystem_action.email_template_new_action_reminder')
-        for action in self:
-            template.send_mail(action.id, force_send=force_send)
+        self.env['mail.template'].browse(template.id).send_mail(
```
grindtildeath: As template is a recordset there's no use of browsing mail.template.
You can simply use : template.send_mail()
Otherwise it makes an unnecessary select in DB.

mgmtsystem_action/models/mgmtsystem_action.py
```diff
@@ -219,5 +209,5 @@ def process_reminder_queue(self, reminder_days=10):
         template = self.env.ref(
             'mgmtsystem_action.action_email_template_reminder_action')
         for action in action_ids:
-            template.send_mail(action.id, force_send=True)
+            self.env['mail.template'].browse(template.id).send_mail(action.id)
```
grindtildeath: Same as before.

grindtildeath (commented):
Just a small improvement, otherwise looks good ! Thanks again for your work.

mgmtsystem_action/models/mgmtsystem_action.py
eugen-don: @grindtildeath Thanks for the spot

Im dooing too much copy paste lately :D

sbidoul (approved):

sbidoul:
Thanks!

ThomasBinsfeld (approved):

eugen-don:
thnaks for your support guys

grindtildeath (approved):

PR #177, opened 2017-03-20
9.0 FIX README.rst

eugen-don:
Hi, there were no more workflows in  mgmtsystem_nonconformity in V. 9.0.
So readme is misleading, since there were no buttons to klick on send for analysis and send for review etc.

Also the Approving users group had no rights and was thus useless, but still in the system and refered in the readme.rst

2017-03-20 10_52_36-management system _ approving user - odoo


dreispt (approved):
Indeed that was has been removed :cry:

max3903 (approved):

eugen-don:
@dreispt 
but we agree we keep this missing logic removed in v10 as well right?

PR #178, opened 2017-03-21
Mgmtsystem nonconformity workflow

eugen-don:
This restores the old behavior from version 8.0 where the quality was ensured by a cycle of analyzing an NC and approving the result of the analysis and using those results as the base for planning measures and deciding on the detailed actions.
In addition, the workflow now supports a new stage.
After the work on the action plan is done, the NC cannot be closed directly. The responsible user has to send his NC to the evaluation stage, where the approving user evaluates the effectiveness of his work on the NC, leaves feedback in the evaluation comments field and has to signs his evaluation. After an approving user signs the evaluation, the responsible user is able to close his case.

mgmtsystem_nonconformity_workflow


dreispt:
IMO this should be done directly on the Nonconformity module, no need to create a separate one.
It should not use the Workflow engine, since it is deprecated.
I didn't review the logic in detail yet, but remember to keep it as light as possible - it easier in an implementation to add additional logic, than to remove the existing one.

Do you agree @max3903 »

max3903:
@dreispt Yes, I agree. 1 pull request per module. Not 2.

eugen-don:
Hi Guys @dreispt  @max3903 

should i then stick to avaid using the workflows and just use the checks like in v 9.0?

dreispt:
@eugen-don Yes. And modify the existing nonconformity module instead of creating a new one.

eugen-don:
@dreispt  
Just in case, I’ll explain the reason for creation of the separate module:

I migrated databases from 8 to 10 for a client. They had complains about the deprecated workflows, and demanded me to port them from 8 to 10, what I then did.

So I was planning on leaving them in the main module, but had already customers who wanted to migrate from 9 to 10 and those were already used to work with the hardcoded checks without the workflows.

Also the fields (analysis_user_id, analysis_date, actions_date, actions_user_id etc.) required for the workflows were gone in v 9, and if allways present would cause trouble in migrations.

And i somehow got the impression that by removing all workflows or hardcoding the workflow of a module into the modules python code, we are taking away some of the freedom the workflow engine provided to users without programming knowledge.

Right now, my customer is able to edit the workflow by adding a field or a button in the views inside the developer tools and adjusting the workflow with the workflow editor like described in this video:
https://www.youtube.com/watch?v=icKKZou-8pg
This requires some technical knowledge but no programming skills in python.
The only disadvantage I had in the past with the workflows was that a database, whose workflows were messed with by the users, were harder to migrate.

And the harder migration process is a good reason for leaving the workflows out.
So ill leave them out in the main module, since they were left in the migration from 8 to 9 anyway.

dreispt:
@eugen-don OK, with that explanation it makes sense to workflows as an option, and that means adding them in a separate module.

While the workflow engine is still there, Odoo intends to "replace" them with simpler linear kanban boards organized in stages. Some minimal consistency logic is hard coded, the the default flow is left free and you should then customize your particular rule, using Automated Actions or even custom code.

The perfect solution is to have common logic in the existing NC module, and then add Wkf specific logic in a separate module.

eugen-don:
@dreispt 
I also think that having common logic in the existing NC module and then adding the Wkf specific logic in a separate one is the way to go.
Note that there already was some wkf logic in the main NC module in V9, but even this was not as complete as in V8, for example in v9 you can skip stage analysis completely… in v8 analysis was required.
So what should I do? Leave the checks from v9 as they were in the NC main module?

dreispt:
In doubt, I'm for alleviating the enforced checks to the minimum.
It's easier to add than to remove logic, both in module improvement and in custom implementation.

PR #181, opened 2017-04-21
[10.0] [IMP] mgmtsystem: document_page_approval as option

ThomasBinsfeld:
- Remove `document_page_approval` from dependencies
- Propose it as an option on `mgmtsystem`

### DIFF
--- a/mgmtsystem/models/res_config.py
+++ b/mgmtsystem/models/res_config.py
@@ -71,6 +71,12 @@ class MgmtsystemConfigSettings(models.TransientModel):
         '-This installs the module mgmtsystem_review.'
     )
 
+    module_document_page_approval = fields.Boolean(
+        'Document page approval',
+        help='Provide document approval and history. \n'
+        '-This installs the module document_page_approval.'
+    )
+
     # Manuals
 
     module_mgmtsystem_quality_manual = fields.Boolean(
--- a/document_page_work_instruction/__manifest__.py
+++ b/document_page_work_instruction/__manifest__.py
@@ -10,8 +10,8 @@
     "license": "AGPL-3",
     "category": "Generic Modules/Others",
     "depends": [
+        'document_page',
         'mgmtsystem',
-        'document_page_approval',
     ],
     "data": [
         'data/document_page.xml',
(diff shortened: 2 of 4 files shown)


max3903:
@ThomasBinsfeld I am not a big fan. Many management system standards require the documentation to be controlled. No control means anyone can make any change at any time and arrange the content to his proper point of view. There should be a referee, someone to manage conflicts.

Let's see what others say.

lasley:
Disclaimer - we're still using Confluence for our mgmt, so take this for a grain of salt due to lack of specific Odoo context. We're planning on migrating to this eventually, hence me commenting.

Due to our org's small size, approvals are superfluous because anyone with access to write also has access to approve. As such, we have explicitly disabled this in our Confluence.

That said, we've obviously had to revert erroneous changes. This is made pretty simple with Confluence's versioning, but I think may be a pain in Odoo. Is there any sort of versioning in this platform?

From a general note - I don't think the ability to enable/disable something could ever be bad, assuming the user knows what they're doing. Is it possible for us to default this as on, with the ability to opt-out via the settings instead?

sbidoul:
@max3903 we have use cases where documents (procedures, manuals, work instructions) are stored in an external system. So all we need in Odoo are URL's of those external documents. The external system takes care of versioning and approval of these documents.

I initially wanted to create a kind of abstract document and let the user choose if he wants document_page or something else. But mgmtsystem is tightly bound to document.page, with the specific difficulty that even a doc category is a document.page. So such an abstraction would be relatively hard to implement and require a complex migration.

So we decided to create a document_page_link addon that lets the user optionally enter a URL as document content.

In such situation, approval and history is superfluous, that's why we'd like to have the choice to not install it.

Our reasoning is that making the installation of the approval module very visible in the Settings section would make it easy enough for the user to discover the feature and decide if he needs it or not.

mgmtsystem/models/res_config.py
```diff
@@ -71,6 +71,12 @@ class MgmtsystemConfigSettings(models.TransientModel):
         '-This installs the module mgmtsystem_review.'
     )
 
+    module_document_page_approval = fields.Boolean(
+        'Document page approval',
+        help='Provide acceptation tools. \n'
```
sbidoul: Provide document approval and history ?

dreispt (approved):

max3903 (approved):

lasley (approved):

PR #169, opened 2017-03-14
Mig mgmtsystem nonconformity to v. 10.0

eugen-don:
Mig mgmtsystem nonconformity to v. 10.0
depends: 
- [x] https://github.com/OCA/management-system/pull/184
- [x] https://github.com/OCA/management-system/pull/168 mgmtsystem action
- [x] https://github.com/OCA/management-system/pull/174 document page procedure

PR #184, opened 2017-04-26
Fix design issue in models/mgmtsystem_nonconformity.py

eugen-don:
preventive_action_id and corrective_action_id are nowhere to be located in the view definition, making it impossible to close a NC that has preventive and corrective actions assignet to it.

ncs_views_v9

### DIFF
--- a/mgmtsystem_nonconformity/tests/test_nonconformity.py
+++ b/mgmtsystem_nonconformity/tests/test_nonconformity.py
@@ -18,9 +18,9 @@ class TestModelNonConformity(common.TransactionCase):
             'description': "description",
             'responsible_user_id': self.env.user.id,
             })
-        action_vals = {'name': 'An Action', 'type_action': 'correction'}
+        action_vals = {'name': 'An Action', 'type_action': 'immediate'}
         action1 = self.nc_model.action_ids.create(action_vals)
-        self.nc_test.corrective_action_id = action1
+        self.nc_test.immediate_action_id = action1
 
     def test_stage_group(self):
         """Group by Stage shows all stages"""
@@ -47,12 +47,15 @@ class TestModelNonConformity(common.TransactionCase):
     def test_done_validation(self):
         """Don't allow closing an NC without evaluation comments"""
         done_stage = self.env.ref('mgmtsystem_nonconformity.stage_done')
+        self.nc_test.action_comments = 'OK!'
         with self.assertRaises(ValidationError):
             self.nc_test.stage_id = done_stage
 
     def test_done_actions_validation(self):
         """Don't allow closing an NC with open actions"""
         done_stage = self.env.ref('mgmtsystem_nonconformity.stage_done')
+        self.nc_test.immediate_action_id.stage_id = \
+            self.env.ref('mgmtsystem_action.stage_open')
         self.nc_test.evaluation_comments = 'OK!'
         with self.assertRaises(ValidationError):
             self.nc_test.stage_id = done_stage
@@ -63,12 +66,14 @@ class TestModelNonConformity(common.TransactionCase):
         self.nc_test.stage_id = self.env.ref(
             'mgmtsystem_nonconformity.stage_open')
         self.assertEqual(
-            self.nc_test.corrective_action_id.stage_id,
+            self.nc_test.immediate_action_id.stage_id,
             self.env.ref('mgmtsystem_action.stage_open'),
             'Plan Approval starts Actions')
 
+        self.nc_test.immediate_action_id.stage_id = \
+            self.env.ref('mgmtsystem_action.stage_open')
         self.nc_test.evaluation_comments = 'OK!'
-        self.nc_test.corrective_action_id.stage_id = \
+        self.nc_test.immediate_action_id.stage_id = \
             self.env.ref('mgmtsystem_action.stage_close')
         self.nc_test.stage_id = self.env.ref(
             'mgmtsystem_nonconformity.stage_done')
--- a/mgmtsystem_nonconformity/demo/mgmtsystem_nonconformity.xml
+++ b/mgmtsystem_nonconformity/demo/mgmtsystem_nonconformity.xml
@@ -2,22 +2,21 @@
 <odoo>
     <data>
 
-        <record id="demo_nonconformity" model="mgmtsystem.nonconformity">
-            <field name="partner_id" ref="base.res_partner_3"/>
-            <field name="create_date">2010-12-03</field>
-            <field name="name">Demo Nonconformity</field>
-            <field name="responsible_user_id" ref="base.user_demo"/>
-            <field name="manager_user_id" ref="base.user_root"/>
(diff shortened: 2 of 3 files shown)


dreispt:
They are in the "Actions" tab.
The "Causes and Analysis" tab only shows the immediate correction actions.

eugen-don:
@dreispt 
i did the following test.
Created new DB with demo enabled, installed NC module and tried to complete demo NC.

The preventive and corrective actions did not show up in actions tab... so i needed to go to actions menu directly search for preventive and corrective actions for this demo NC and set them to closed, in order to complete the demo NC (set it to closed)... Did i miss something in this workflow?

eugen-don:
seems like this problem occurs only on demo data

dreispt:
Even if the problem is from demo data, a fix is needed.
I tested on Runbot, and I get the same message.
But the Action Plan is empty, and the Immediate Action is done.
Probably there is a bug at https://github.com/OCA/management-system/blob/9.0/mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py#L203 ...

Con you open an issue or, even better, try to find a fix?

eugen-don:
@dreispt 
Ill try fixing it first, before opening any issues

dreispt:
Great! The point for that validation is to allow closing an NC if there is any Action pending (Immediate, or not).

eugen-don:
@dreispt 

I looked at the model and here is what I think is the problem: 
In mgmtsystem_nonconformity.py we have 1 field (action_ids) with a Many2many relation 
and 3 fields (immediate_action_id, corrective_action_id and preventive_action_id) with a Many2one relation, all to the same model (mgmtsystem.action).

The names of the 3 fields (immediate_action_id, corrective_action_id and preventive_action_id) indicate their intended purpose, it was to select the specific response type.

The field type_action is a selection field with 4 values and there is no need to have a relation field for each of the values if the intention is just to get a record with a specific value.
Domain filters are the way to go wether in py code or view definitions, although views are more flexible to update/extend and faster.

Oddly is also the fact that we already use a filter in the analysis tab where the immediate action is shown, but corrective_action_id, preventive_action_id as well as opportunities are nowere declared in the views.

Since the field action_ids is a Many2many relation it includes all actions that are created inside the form view of the NC in the actions tab regardless of the actions type.

But if you import data or demo data is created where only the relation for the many2one fields (immediate_action_id, corrective_action_id and preventive_action_id) is set and the Many2many relation of action_ids is skipped,
we recieve the error that some actions arent done since the Many2many field action_ids is not displaying them.

Now a possible solution is to add the Many2one fields unter the action_ids field in the actions tab. This is the easiest solution, but this would lead Odoo displaying the actions multiple times.

A better solution though IMO is to remove the Many2one fields and change the views accordingly to use the Many2many field instead with a domain like the example below

`<field name="immediate_action_id" domain="[('type_action','=','immediate')]" attrs="{'readonly':[('state','not in',['analysis'])]}"/>

<field name="action_ids" domain="[('type_action','=','immediate')]" attrs="{'readonly':[('state','not in',['analysis'])]}"/>
`
Then all complicated checks for the Many2one fields could also be removed... some less code to carry around...

If i didnt miss any essential meaning i would like to try and smallen down this concept by removing the M2o fields.
Although if there is any other way, like to compute the Many2many and Many2one fields together (i know this works but just makes the module bigger), or maybe a way to have the views manage this conditionally, id also agree to try and solve it this way...

Any advice on how i should go about this? Or am i lost completely?

lasley:
@eugen-don - The issue lies [in these three lines of demo](https://github.com/eugen-don/management-system/blob/203984692c5770a13e3ba47940ef0fbb0ef19461/mgmtsystem_nonconformity/demo/mgmtsystem_nonconformity.xml#L17-L19). If you look at [the declaration for them](https://github.com/eugen-don/management-system/blob/203984692c5770a13e3ba47940ef0fbb0ef19461/mgmtsystem_action/demo/mgmtsystem_action.xml) you'll note that there is no state, therefore they are defaulting to open.

Searching the error in the codebase leads me [here](https://github.com/OCA/management-system/blob/711c0e7932d01d25db67a821ad3b1c3df2359279/mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py#L202), which indicates that if the noncomformity is being marked as done, all of its actions must be closed.

Have you closed all three actions associated with this nonconformity?

dreispt:
There is design issue: I believe that the `corrective_action_id` and `preventive_action_id` were deprecated and replaced by the `action_ids` field.

The fix is change the demo data to use the `action_ids` field instead, and to remove the deprecated fields.

max3903:
@dreispt You are exactly right.

eugen-don:
@dreispt @max3903 
Hi guys
i was able to make the demo functional by replacing the records for the corrective_action_id and preventive_action_id with a normal action_ids record.

But when removing the unused fields corrective_action_id  and preventive_action_id  i discovered some tests that were written for them. I wasnt able to get the tests passing, and also codecoverage would drop if i removed them.

Anyways the demo should work fine now, since there are no more hidden open actions behind a NC.

Now just in case ill ask, is it ok to remove tests and have a decrease in codecoverage? Or do you guys have some other way of handling such cases ?

lasley:
@eugen-don - I'm unclear on your question. Tests look to be passing & there is no drop in coverage.

max3903:
@eugen-don Tests using deprecated fields should be deprecated as well and fields and tests should even be removed. 

It would be nice to have new tests on the new fields though.

dreispt:
Ideally existing tests would be adapted to use the new field rather than the old ones.

eugen-don:
OK ill modify the tests to use the new fields instead of the old ones and remove and remove the old fields from the code in the model.

mgmtsystem_nonconformity/tests/test_nonconformity.py
```diff
@@ -18,9 +18,9 @@ def setUp(self):
             'description': "description",
             'responsible_user_id': self.env.user.id,
             })
-        action_vals = {'name': 'An Action', 'type_action': 'correction'}
+        action_vals = {'name': 'An Action', 'type_action': 'immediate'}
         action1 = self.nc_model.action_ids.create(action_vals)
-        self.nc_test.corrective_action_id = action1
+        self.nc_test.immediate_action_id = action1
```
eugen-don: simply switching from corrective_action_id to immediate_action_id

eugen-don:
Sorry for the wait guys...

now this seems to solve the error in the demo...

eugen-don:
Ping @dreispt @max3903 @lasley

angelmoya (approved):

max3903:
@eugen-don is it still wip?

dreispt (approved):

eugen-don:
@max3903 
no its ready

eugen-don:
Since this is merged in 9.0 the PR for MIG https://github.com/OCA/management-system/pull/169
will be WIP once more to account for the changes made here... we dont want the same issue in 10.0

PR #188, opened 2017-06-21
[ADD] Module which add processus template

benwillig:
Add a processus template in management system documentation.


max3903:
Hello @benwillig 

Why not using the procedure module?

benwillig:
@max3903 Procedure and processes are not the same thing. But I understand that using the same template could be confusing.

dreispt:
I think Maxime means that the `document_page_procedure` should be extended to provide both Procedure and Process templates. I agree with this point of view - no need to have two separate modules for that.

PR #192, opened 2017-07-11
[10.0][FIX] Causes selection in a nonconformity

benwillig:
A simple management user has no create/delete rights on nonconformity causes. Then, he is not able to select/deselect causes on a nonconformity.

### DIFF
--- a/mgmtsystem_nonconformity/views/mgmtsystem_nonconformity.xml
+++ b/mgmtsystem_nonconformity/views/mgmtsystem_nonconformity.xml
@@ -158,7 +158,15 @@
                 <field name="analysis" attrs="{'readonly':[('state','not in',['analysis'])]}"/>
 
                 <separator string="Causes"/>
-                <field name="cause_ids" attrs="{'readonly':[('state','not in',['analysis'])]}"/>
+                <field name="cause_ids" attrs="{'readonly':[('state','not in',['analysis'])]}"
+                       options="{'no_create': True}">
+                  <tree create="1" delete="1" colors="blue:parent_id;">
+                    <field name="parent_id"/>
+                    <field name="name"/>
+                    <field name="description"/>
+                    <field name="sequence"/>
+                  </tree>
+                </field>
 
                 <separator string="Analysis Confirmation"/>
                 <group>


dreispt:
Sorry, I don't understand the point for this change. Can you explain?

benwillig:
There is a bug in v10 (which won't be fix by Odoo). For example, you have a model like following.

```python
class Test(models.Model):
    _name = 'test'
    user_ids = fields.Many2many(comodel_name='res.users')
```
Then you have a view with following architecture.

```xml
<form>
    <field name="user_ids"/>
</form>
```

Users who have the create and delete rights on ```res.users``` won't have any problems.

A simple user (w/o create and delete access) will not be able to select users, because he has no rights to create it. And he won't be able to deselect items, because he has no delete rights. The only workaround is to force ```create=True``` and ```delete=True``` in field architecture.

See https://github.com/odoo-dev/odoo/commit/0a157e599999fe4d93694a444d0a6fa0e6940379

dreispt:
Oh my! 😱 
I need to double check that.

eugen-don:
Hi @benwillig 
sorry gues i missunder stood the issue.

I just veryfied this behaviour.
But the desired workflow is to have the analysis be done by a user in the management-systems manager group.

benwillig:
@eugen-don I never said that a Management system user can't create nonconformities. The purpose of this fix is to allow a basic management system user to select/deselect causes on a nonconformity. He has no rights to create/delete ```mgmtsystem.nonconformity.cause```. The example I gave explained what is the bug in Odoo.

image

eugen-don:
Hi @benwillig 

yes, i just misread some other issue.

I confirmed the issue, its an odoo bug and needs fixing.

Though i dont know how long it wil take for them to fix this.

@dreispt 
we should consider merging this, since its a clean workaround for the odoo bug.

eugen-don (approved):

benwillig:
I already opened a ticket about that, and they said it won't be fix in 10.0 because it needs a deep refactoring. But maybe you will be luckier than me, so I guess it would be a good idea to open an other ticket.

eugen-don:
@benwillig thanks again for the workaround.

I took a look at your fix and i looks easy and clean, i doubt there is any deeper refactoring needed.
Anyways ill try my luck as well just in case...

eugen-don:
Hi @dreispt 
i did some extesive testing.
Everything works as expected and i did not discover any security issues.
I think we can merge this savely.

dreispt (approved):

PR #193, opened 2017-07-11
[10.0][FIX] Auditor/Auditee selection in an audit

benwillig:
A manager has no create/delete rights on users. Then, he is not able to select/deselect users on an audit.

### DIFF
--- a/mgmtsystem_audit/views/res_users.xml
+++ b/mgmtsystem_audit/views/res_users.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2017 ACSONE SA/NV
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+
+    <record model="ir.ui.view" id="mgmtsystem_audit_res_users_x2m_tree">
+        <field name="name">mgmtsystem.audit.res.users.x2m.tree (in mgmtsystem_audit)</field>
+        <field name="model">res.users</field>
+        <field name="arch" type="xml">
+            <tree create="1" delete="1">
+                <field name="name"/>
+                <field name="login"/>
+                <field name="lang"/>
+                <field name="login_date"/>
+            </tree>
+        </field>
+    </record>
+
+</odoo>
--- a/mgmtsystem_audit/__manifest__.py
+++ b/mgmtsystem_audit/__manifest__.py
@@ -3,7 +3,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {
     "name": "Management System - Audit",
-    "version": "10.0.1.0.0",
+    "version": "10.0.1.0.1",
     "author": "Savoir-faire Linux, Odoo Community Association (OCA)",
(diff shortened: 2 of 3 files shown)


mgmtsystem_audit/views/res_users.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2017 ACSONE SA/NV
```
eugen-don, line 2: could you remove the Copyright from the xml?
Its uncommon to copyright view definitions.

sbidoul: @eugen-don [copyright in the views is part of the OCA template](https://github.com/OCA/maintainer-tools/blob/4679ce184af4d2af3175d0b52f3178b0c231b83c/template/module/views/model_name_view.xml#L2-L3).

eugen-don:
@benwillig 
did you actually test this?
Pls, take one more look at the audit module and the settings

mgmtsystem_audit/views/res_users.xml
eugen-don: @sbidoul 
sorry i didnt notice yet

eugen-don:
@benwillig 
you need to make sure the user is actually in the managers / auditors group ? or am i missing something?

Sorry about the confusion, but you use the same view ref twice (mgmtsystem_audit_res_users_x2m_tree), meaning auditors and auditees would be the same people... it there any way you could fix this ?

benwillig:
I used the wrong view ref key indeed, I fixed it. ```form_view_ref``` and ```tree_view_ref``` key allow to open /show a specific view using its external id.

eugen-don:
@benwillig 
nice thanks

eugen-don:
@benwillig 
I did some testing and everything looks nice and works as expected.

The only litle thing i dont like is the create button, thats still displayed even though the user has no create rights on res.users
image
Any way to get it the same way as in the Nonconformities module?

If this is to much work, id still aprove since the buttons doesnt do anything by itself.

benwillig:
@eugen-don I added the option.

eugen-don:
Hi @benwillig 
somehow now the add buttons gone completely? Maybe its just a local cache issue on my machine?
Can you check it with an manager account just in case?

eugen-don:
@benwillig 
sorry.. it was my cache... Everything works fine!

eugen-don:
Hi @dreispt 
This is a fix for the same issue as https://github.com/OCA/management-system/pull/192 , we can merge it savely.

eugen-don (approved):

ThomasBinsfeld (approved):
Code and functional review :+1:

zakiuu (approved):
Code looks good, I guess you can squash the commits

benwillig:
Commits squashed!

rami-wafaie (approved):
code review and testing :+1:

pedrobaeza:
Can you change module version number?

benwillig:
Done!

adrienpeiffer (approved):
:+1: Ready to merge

PR #201, opened 2017-10-26
[FIX] Set create_date as readonly because it can't be changed anyway

benwillig:
```create_date``` can't be edited. If we want an editable date field, we should add another one.


adrienpeiffer (approved):
:+1:

ivantodorovich (approved):
LGTM 👍

max3903 (approved):

Issue #202, opened 2017-11-07
It should be 'web.base.url' (in our systems the url link in the action email always points to localhost:8069


fdvarela:
https://github.com/OCA/management-system/blob/a4f747ff21abf092b776d6602543e06d40b198ba/mgmtsystem_action/mgmtsystem_action.py#L97

max3903:
Hello @fdvarela 

Can you provide your version of Odoo please?

Thanks!

fdvarela:
Version 8

fdvarela:
Also the URL generated is weird. 
To workaround the issue I created web.base_url (with underscore) in the configuration and set it to our URL.
Now the URL in the email (the link to the action) is (HTML extract)

< a href = "https://erp.monteloeder.com/%5Bu'https://erp.monteloeder.com?db=monteloeder#model=mgmtsystem.action&amp;id=1'%5D" style="font-weight: bold;">.....< / a >

It seems the <a href tag is prepended with the web.base_url and it is concatenated with the full (correct) url

PR #203, opened 2017-11-07
[FIX] Issue #202

max3903:
[FIX] Issue #202

### DIFF
--- a/mgmtsystem_action/__openerp__.py
+++ b/mgmtsystem_action/__openerp__.py
@@ -20,7 +20,7 @@
 ##############################################################################
 {
     "name": "Management System - Action",
-    "version": "8.0.1.2.2",
+    "version": "8.0.1.2.3",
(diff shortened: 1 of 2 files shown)


max3903:
@fdvarela Please review the fix.

fdvarela:
It won't suffice. 

The email's html body is passed through a function defined in *email_template/email_template.py* line 131 
def _replace_local_links(self, cr, uid, html, context=None)

which replaces "img src" and "a href" urls prepending *web.base.url* (the right one)

I think the fix in the management system should only build the "path" part of the url and let the email_template function prepend the base url.

dreispt (approved):
LGTM

ivantodorovich (approved):
LGTM, although automated checks are not running.

ivantodorovich:
Shouldn't this be ported to 9.0 and 10.0?

Issue #200, opened 2017-10-03
Migration to version 11.0


pedrobaeza:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

# Modules to migrate

Dependencies on Knowledge:
- [x] https://github.com/OCA/knowledge/issues/147

## Quality

- [x] document_page_procedure by @jalzaga - #214 
- [x] document_page_quality_manual by @jalzaga - #216 
- [x] document_page_work_instruction by @jalzaga - #213 
- [x] mgmtsystem by @jalzaga - #210 
- [x] mgmtsystem_action by @jalzaga - #211 
- [x] mgmtsystem_audit by @jalzaga - #220 
- [x] mgmtsystem_manual by @jalzaga - #215 
- [x] mgmtsystem_nonconformity by @jalzaga - #217 
- [x] mgmtsystem_quality by @jalzaga - #221 
- [x] mgmtsystem_review by @jalzaga - #219 
- [x] mgmtsystem_survey by @jalzaga - #218 

## Environment

- [ ] document_page_environment_manual
- [ ] document_page_environmental_aspect
- [ ] mgmtsystem_environment

## Health & Safety

- [ ] document_page_health_safety_manual
- [ ] mgmtsystem_hazard
- [ ] mgmtsystem_health_safety

## Information Security

- [ ] mgmtsystem_info_security_manual

## Advanced Features

- [ ] mgmtsystem_action_audittrail
- [ ] mgmtsystem_audit_audittrail
- [ ] mgmtsystem_claim
- [ ] mgmtsystem_claim_audittrail
- [ ] mgmtsystem_hazard_audittrail
- [ ] mgmtsystem_kpi
- [ ] mgmtsystem_nonconformity_analytic
- [ ] mgmtsystem_nonconformity_claim
- [ ] mgmtsystem_nonconformity_deptm
- [ ] mgmtsystem_nonconformity_project
- [ ] mgmtsystem_review_audittrail

jalzaga:
Is there a PR for it? we´ll be ready to test as soon as it is available, please inform us if you are going to assign works to other contributors to participate.

max3903:
@jalzaga There is no contributors to work on this yet. If you have resources available (human or financial), please let me know and we can coordinate the effort.

jalzaga:
We can dedicate some time from our resources, both human and financial, take us into account. @max3903

jalzaga:
[...]

PR #213, opened 2018-04-15
[11.0][MIG] document page work instruction

jalzaga:
Module document_page_work_instruction Migration to version 11.0 #200


etobella:
@jalzaga Can you rebase it in order to review it?

etobella:
Missing dependencies on OCA/knowledge. Add it on oca_dependencies.txt

jalzaga:
This PR is Ready to test

document_page_work_instruction/views/document_page_work_instructions.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+
+    <record model="ir.actions.act_window" id="work_instructions_action">
+      <field name="name">Work Instructions</field>
+      <field name="res_model">document.page</field>
+      <field name="view_type">form</field>
+      <field name="view_mode">tree,form</field>
+      <field name="help">Work instructions of your management system.</field>
+      <field name="domain" eval="[('parent_id','=',ref('document_page_group_work_instructions'))]"/>
```
etobella, line 10: I don't know your opinion, but I think that managing the category of the work instruction through the parent is something fishy... 
I think it would be better to manage it through an inherits or a boolean invisible field, I don't know what do you think, but this kind of management could give problems (users may change the category and the work instructions disappear...).
The change should be managed on a migration script.

etobella (commented):
Just a few comments.

document_page_work_instruction/views/document_page_work_instructions.xml
max3903: @etobella What about preventing the user to change the category?

max3903 (approved):

ivantodorovich (changes requested):
remove old migration scripts, and possibly squash your 3 migration commits

document_page_work_instruction/views/document_page_work_instructions.xml
ivantodorovich: Although I don't have a formed opinion on this, it's been like this since previous versions. Maybe this should be a separate PR as an IMP.

jalzaga:
@ivantodorovich shall I remove the whole directory migrations/7.0.1.0 or just the pre-migration.py?

document_page_work_instruction/views/document_page_work_instructions.xml
etobella: I think that prevent the user to change the category should be a good way to fix this problem.

ivantodorovich:
> @ivantodorovich shall I remove the whole directory migrations/7.0.1.0 or just the pre-migration.py?

Remove the whole directory.

ivantodorovich (approved):

document_page_work_instruction/views/document_page_work_instructions.xml
pedrobaeza: Is this going to be attended in this PR or on a new one?

max3903: @ivantodorovich Can you create an issue so we can have a discussion separately on this topic and get this pull request merged?

ivantodorovich: Follow up here: https://github.com/OCA/management-system/issues/237

PR #215, opened 2018-04-15
[11.0][MIG] mgmtsystem manual

jalzaga:
Module mgmtsystem_manual Migration to version 11.0 #200


etobella:
@jalzaga Can you rebase it in order to review it?

etobella:
Missing dependency on OCA/knowledge. Can you add it on oca_dependencies?

max3903:
@jalzaga Same issue as in #218. See my comment there about rebasing.

mgmtsystem_manual/models/mgmtsystem_manual.py
```diff
@@ -0,0 +1,9 @@
+
+from odoo import models, fields
+
+
+class MgmtSystemManual(models.Model):
+
+    _inherit = 'mgmtsystem.system'
+    _description = 'Manual'
+    manual = fields.Many2one('document.page', string='Manual')
```
etobella: You should add a domain, because all document pages can be selected

etobella (changes requested):
You should fix these error in order to make runbot available.

mgmtsystem_manual/models/mgmtsystem_manual.py
```diff
@@ -0,0 +1,9 @@
+
+from odoo import models, fields
+
+
+class MgmtSystemManual(models.Model):
+
+    _inherit = 'mgmtsystem.system'
+    _description = 'Manual'
```
etobella, line 8: Why you modify the description of mgmtsystem.system?

mgmtsystem_manual/models/mgmtsystem_manual.py
etobella: It should be renamed as manual_id. You should include that manual was the oldname.

And just a question. Why are you making it as part of a category? It can be changed...
Maybe, you should add a boolean like: is_mgmsystem_manual. It's default is false, but you add on the context that it is true on this view.

etobella (commented):
Just a few comments that I found.

jalzaga:
This lines come from the v10, they haven´t been changed, however we´ll make the corrections you consider necessary. @etobella

jalzaga:
This PR is ready to test

etobella:
I see the same problem that I showed on #213, but as I said, it's just my opinion.

max3903 (approved):

mgmtsystem_manual/migrations/7.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,65 @@
+##############################################################################
```
ivantodorovich: Old migration scripts should be removed

mgmtsystem_manual/models/mgmtsystem_manual.py
```diff
@@ -6,4 +6,5 @@ class MgmtSystemManual(models.Model):
 
     _inherit = 'mgmtsystem.system'
     _description = 'Manual'
-    manual = fields.Many2one('document.page', string='Manual')
+    manual_id = fields.Many2one('document.page', string='Manual')
+# old name of manual_id was manual
```
ivantodorovich: shouldn't you use `oldname='manual'` to help with migration?
The comment line is unnecessary

ivantodorovich (changes requested):

mgmtsystem_manual/models/mgmtsystem_manual.py
max3903: Agree

mgmtsystem_manual/models/mgmtsystem_manual.py
```diff
@@ -7,4 +7,4 @@ class MgmtSystemManual(models.Model):
     _inherit = 'mgmtsystem.system'
     _description = 'Manual'
     manual_id = fields.Many2one('document.page', string='Manual')
-# old name of manual_id was manual
+    oldname = 'manual'
```
ivantodorovich: It should be like this:
```py
manual_id = fields.Many2one('document.page', string='Manual', oldname='manual')
```

ivantodorovich (changes requested):

ivantodorovich (approved):

pedrobaeza:
Only one little PEP8 fail:
```
mgmtsystem_manual/models/mgmtsystem_manual.py:9:80: E501 line too long (83 > 79 characters)
```

PR #226, opened 2018-06-28
[ADD] mgmtsystem_consent: Ask for consent for data processing

yajo:
This addon makes it possible to define which partners are included
in any management system, and to define an automated explicit consent
acceptance system.

I'm not sure if this should go here or in OCA/data-protection, since it has much to do with both places... :thinking: 

Well, publishing here for now.

@Tecnativa


mgmtsystem_consent/models/res_partner.py
```diff
@@ -0,0 +1,26 @@
...
+        "Consents",
+        compute="_compute_mgmtsystem_consent_ids_count",
+        help="Management system consent requests amount",
+    )
+
+    @api.depends("mgmtsystem_consent_ids")
+    def _compute_mgmtsystem_consent_ids_count(self):
+        """Count consent requests."""
+        for one in self:
+            one.mgmtsystem_consent_ids_count = len(one.mgmtsystem_consent_ids)
```
pedrobaeza, line 26: Remember to use `read_group` for not impacting too much on performance.

mgmtsystem_consent/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+This module aims for GDPR compliance, enabling the user to define a set of
```
pedrobaeza, line 1: Put a more generic description, although the GDPR example can be also here.

pedrobaeza (changes requested):
Seeing the content, I think this should go on OCA/data-protection, although depending on mgmt_system

You should add some basic tests, avoiding mocks as possible, to be sure about the features.

dreispt:
There are many managerial related aspects of "Data privacy and protection" that are better modeled as a "Management System". It is a grey area but I tend to agree with Pedro, and prefer to add that to the data-privacy repo.

More specifically on the features added, it claims to be a consent record, rather than an actual process/procedure management system, so that left me confused.

Also the description states that we are storing consent for people to be included in management systems. For the use cases I'm familiar with this does not make sense: these systems are integral part of my operation, maybe even mandatory by industry or government regulation, and people can't opt out.

Maybe a more specific examples could help understanding the use case this feature is targeting.

yajo:
Before you lose more time reviewing, I'm then gonna narrow this PR here to just the modifications to the base mgmtsystem addon I need, and then open a new one in data-protection for the new mgmtsystem_consent one.

Somebody can set as WIP.

yajo:
After some insightful conversation with @pedrobaeza, this is gonna fully land in OCA/data-protection.

PR #227, opened 2018-07-04
Make management system available in V11

jguerriat:
Some minors changes have been made to make management system available for V11


etobella:
You should make a PR for each module. However, Some PR have been done, so you could review them.
Moreover, you added some duplicated modules. For example, document_page can be found on https://github.com/OCA/knowledge
Can you review if all the proposals that you made are on the other PRs?

etobella:
It seems that this issue is no longer open because we have several PR for each module.
@pedrobaeza can you close it?

PR #241, opened 2019-02-19
mgmtsystem: Access rights for document pages

ernestotejeda:
It gives right access to document.page of management system to the users of the 'Management System' groups.
With this, it's not necessary for the users of the Management System to be added to a Knowledge group in order to have access to their document.page, since adding them to a Knowledge group implies that they see all the document.pages and also everything under the Knowledge menu

Cc @Tecnativa

### DIFF
--- a/document_page_environmental_aspect/__manifest__.py
+++ b/document_page_environmental_aspect/__manifest__.py
@@ -20,7 +20,7 @@
 ##############################################################################
 {
     "name": "Environmental Aspects",
-    "version": "9.0.1.0.0",
+    "version": "10.0.1.0.0",
     "author": "Savoir-faire Linux, Odoo Community Association (OCA)",
     "website": "http://www.savoirfairelinux.com",
     "license": "AGPL-3",
@@ -32,6 +32,7 @@
     "data": [
         'data/document_page.xml',
         'views/document_page.xml',
+        'security/environmental_aspect_security.xml',
     ],
-    'installable': False,
+    'installable': True,
 }
--- a/document_page_environmental_aspect/security/environmental_aspect_security.xml
+++ b/document_page_environmental_aspect/security/environmental_aspect_security.xml
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <!-- Read/Write/Create/Unlink access in 'Management System / viewer' group
+        to 'Environmental aspect' documents -->
+    <record id="environmental_document_rule_mgmtsystem_viewer" model="ir.rule">
+        <field name="name">Environmental document page for viewers</field>
+        <field name="model_id" ref="document_page.model_document_page"/>
+        <field name="domain_force" eval="['|', ('parent_id','=',ref('document_page_group_environmental_aspect')), ('id','=',ref('document_page_group_environmental_aspect'))]"/>
+        <field name="groups" eval="[(4, ref('mgmtsystem.group_mgmtsystem_viewer'))]"/>
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
+    </record>
+
+    <record id="environmental_document_history_rule_mgmtsystem_viewer" model="ir.rule">
+        <field name="name">Environmental document page for viewers</field>
+        <field name="model_id" ref="document_page.model_document_page_history"/>
+        <field name="domain_force" eval="['|', ('page_id.parent_id','=',ref('document_page_group_environmental_aspect')), ('page_id','=',ref('document_page_group_environmental_aspect'))]"/>
+        <field name="groups" eval="[(4, ref('mgmtsystem.group_mgmtsystem_viewer'))]"/>
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
+    </record>
+</odoo>
--- a/document_page_procedure/security/procedure_security.xml
+++ b/document_page_procedure/security/procedure_security.xml
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <!-- Read/Write/Create/Unlink access in 'Management System / viewer' group
+        to 'Procedure' documents -->
+    <record id="procedure_document_rule_mgmtsystem_viewer" model="ir.rule">
+        <field name="name">Procedure document page for viewers</field>
+        <field name="model_id" ref="document_page.model_document_page"/>
+        <field name="domain_force" eval="['|', ('parent_id','=',ref('document_page_group_procedure')), ('id','=',ref('document_page_group_procedure'))]"/>
+        <field name="groups" eval="[(4, ref('mgmtsystem.group_mgmtsystem_viewer'))]"/>
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
+    </record>
+
+    <record id="procedure_document_history_rule_mgmtsystem_viewer" model="ir.rule">
+        <field name="name">Procedure document page for viewers</field>
+        <field name="model_id" ref="document_page.model_document_page_history"/>
+        <field name="domain_force" eval="['|', ('page_id.parent_id','=',ref('document_page_group_procedure')), ('page_id','=',ref('document_page_group_procedure'))]"/>
+        <field name="groups" eval="[(4, ref('mgmtsystem.group_mgmtsystem_viewer'))]"/>
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
+    </record>
+</odoo>
--- a/document_page_work_instruction/security/work_instruction_security.xml
+++ b/document_page_work_instruction/security/work_instruction_security.xml
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <!-- Read/Write/Create/Unlink access in 'Management System / viewer' group
+        to 'Work instruction' documents -->
+    <record id="work_document_rule_mgmtsystem_viewer" model="ir.rule">
+        <field name="name">Work document page for viewers</field>
+        <field name="model_id" ref="document_page.model_document_page"/>
+        <field name="domain_force" eval="['|', ('parent_id','=',ref('document_page_group_work_instructions')), ('id','=',ref('document_page_group_work_instructions'))]"/>
+        <field name="groups" eval="[(4, ref('mgmtsystem.group_mgmtsystem_viewer'))]"/>
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
+    </record>
+
+    <record id="work_document_history_rule_mgmtsystem_viewer" model="ir.rule">
+        <field name="name">Work document page history for viewers</field>
+        <field name="model_id" ref="document_page.model_document_page_history"/>
+        <field name="domain_force" eval="['|', ('page_id.parent_id','=',ref('document_page_group_work_instructions')), ('page_id','=',ref('document_page_group_work_instructions'))]"/>
+        <field name="groups" eval="[(4, ref('mgmtsystem.group_mgmtsystem_viewer'))]"/>
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
+    </record>
+</odoo>
--- a/mgmtsystem/security/ir.model.access.csv
+++ b/mgmtsystem/security/ir.model.access.csv
@@ -1,5 +1,10 @@
 "id","name","model_id:id","group_id:id","perm_read","perm_write","perm_create","perm_unlink"
 "access_mgmtsystem_system_user","mgmtsystem.system.user","model_mgmtsystem_system","mgmtsystem.group_mgmtsystem_user",1,0,0,0
 "access_mgmtsystem_system_viewer","mgmtsystem.system.viewer","model_mgmtsystem_system","mgmtsystem.group_mgmtsystem_viewer",1,0,0,0
+"access_document_page_mgmtsystem_viewer","document.page.mgmtsystem.viewer","document_page.model_document_page","group_mgmtsystem_viewer",1,0,0,0
+"access_document_page_history_mgmtsystem_viewer","document.page.history.mgmtsystem.viewer","document_page.model_document_page_history","group_mgmtsystem_viewer",1,0,0,0
 "access_mgmtsystem_system_auditor","mgmtsystem.system.auditor","model_mgmtsystem_system","mgmtsystem.group_mgmtsystem_auditor",1,0,0,0
+"access_document_page_mgmtsystem_auditor","document.page.mgmtsystem.auditor","document_page.model_document_page","group_mgmtsystem_auditor",1,1,1,0
+"access_document_page_history_mgmtsystem_auditor","document.page.history.mgmtsystem.auditor","document_page.model_document_page_history","group_mgmtsystem_auditor",1,1,1,0
 "access_mgmtsystem_system_manager","mgmtsystem.system.manager","model_mgmtsystem_system","mgmtsystem.group_mgmtsystem_manager",1,1,1,1
+"access_document_page_mgmtsystem_manager","document.page.mgmtsystem.manager","document_page.model_document_page","group_mgmtsystem_manager",1,1,1,1
--- a/mgmtsystem/security/mgmtsystem_security.xml
+++ b/mgmtsystem/security/mgmtsystem_security.xml
@@ -20,14 +20,6 @@
         <field name="implied_ids" eval="[(4,ref('mgmtsystem.group_mgmtsystem_viewer'))]"/>
     </record>
 
-    <record id="group_mgmtsystem_user_manager" model="res.groups">
-        <field name="name">Approving User</field>
-        <field name="comment">Module user with additional permission for approvals.</field>
-        <field name="category_id" ref="module_category_management_system"/>
-        <field name="implied_ids" eval="[(4,ref('mgmtsystem.group_mgmtsystem_user'))]"/>
-        <field name="users" eval="[(6,0,[ref('base.user_root')])]"/>
-    </record>
-
     <record id="group_mgmtsystem_auditor" model="res.groups">
         <field name="name">Auditor</field>
         <field name="comment">Module user with additional permission specific for the auditor job, such as effectiveness reviews.</field>
@@ -52,4 +44,49 @@
         <field name="domain_force">['|',('company_id','=',False),('company_id','child_of',[user.company_id.id])]</field>
     </record>
 
+    <!-- No page access by default; submodules add access to specific sections -->
+    <record id="no_document_rule_mgmtsystem_viewer" model="ir.rule">
+        <field name="name">No document page for viewers</field>
+        <field name="model_id" ref="document_page.model_document_page"/>
+        <field name="domain_force" eval="[(0, '=', 1)]"/>
+        <field name="groups" eval="[(4, ref('mgmtsystem.group_mgmtsystem_viewer'))]"/>
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
+    </record>
+
+    <record id="no_document_history_rule_mgmtsystem_viewer" model="ir.rule">
+        <field name="name">No document page history for viewers</field>
+        <field name="model_id" ref="document_page.model_document_page_history"/>
+        <field name="domain_force" eval="[(0, '=', 1)]"/>
+        <field name="groups" eval="[(4, ref('mgmtsystem.group_mgmtsystem_viewer'))]"/>
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
+    </record>
+
+    <!-- Read access in 'Knowledge / Knowledge User' group for all documents -->
+    <record id="document_page_rule_document_user_r" model="ir.rule">
+        <field name="name">All document pages (read)</field>
+        <field name="model_id" ref="document_page.model_document_page"/>
+        <field name="domain_force">[(1, '=', 1)]</field>
+        <field name="groups" eval="[(4, ref('knowledge.group_document_user'))]"/>
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
+    </record>
+
+    <record id="document_page_history_rule_document_user_r" model="ir.rule">
+        <field name="name">All document pages history (read)</field>
+        <field name="model_id" ref="document_page.model_document_page_history"/>
+        <field name="domain_force">[(1, '=', 1)]</field>
+        <field name="groups" eval="[(4, ref('knowledge.group_document_user'))]"/>
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
+    </record>
 </odoo>
--- a/mgmtsystem_manual/security/manual_security.xml
+++ b/mgmtsystem_manual/security/manual_security.xml
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <!-- Read/Write/Create/Unlink access in 'Management System / viewer' group
+        to 'Manual' documents -->
+    <record id="manual_document_rule_mgmtsystem_viewer" model="ir.rule">
+        <field name="name">Manual document page for viewers</field>
+        <field name="model_id" ref="document_page.model_document_page"/>
+        <field name="domain_force" eval="['|', ('parent_id','ilike', 'manual'), ('name','ilike', 'manual')]"/>
+        <field name="groups" eval="[(4, ref('mgmtsystem.group_mgmtsystem_viewer'))]"/>
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
+    </record>
+
+    <record id="manual_document_history_rule_mgmtsystem_viewer" model="ir.rule">
+        <field name="name">Manual document page for viewers</field>
+        <field name="model_id" ref="document_page.model_document_page_history"/>
+        <field name="domain_force" eval="['|', ('page_id.parent_id','ilike', 'manual'), ('page_id.name','ilike', 'manual')]"/>
+        <field name="groups" eval="[(4, ref('mgmtsystem.group_mgmtsystem_viewer'))]"/>
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
+    </record>
+</odoo>
--- a/document_page_procedure/__manifest__.py
+++ b/document_page_procedure/__manifest__.py
@@ -15,6 +15,7 @@
     "data": [
         'data/document_page_procedure.xml',
         'views/document_page_procedure.xml',
+        'security/procedure_security.xml',
     ],
     "demo": [
         'demo/document_page_procedure.xml',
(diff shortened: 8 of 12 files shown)


yajo:
Could you explain this please?

ernestotejeda:
> It gives right access to document.page of management system to the users of the 'Management System' groups.
> With this, it's not necessary for the users of the Management System to be added to a Knowledge group in order to have access to their document.page, since adding them to a Knowledge group implies that they see all the document.pages and also everything under the Knowledge menu

@Yajo, Now I wrote a description in the first comment that explains the purpose of this PR.

mgmtsystem/security/ir.model.access.csv
```diff
@@ -1,5 +1,13 @@
...
+"access_document_page_mgmtsystem_viewer","document.page.mgmtsystem.viewer","model_document_page","mgmtsystem.group_mgmtsystem_viewer",1,0,0,0
+"access_document_page_history_mgmtsystem_viewer","document.page.history.mgmtsystem.viewer","document_page.model_document_page_history","mgmtsystem.group_mgmtsystem_viewer",1,0,0,0
 "access_mgmtsystem_system_auditor","mgmtsystem.system.auditor","model_mgmtsystem_system","mgmtsystem.group_mgmtsystem_auditor",1,0,0,0
+"access_document_page_mgmtsystem_auditor","document.page.mgmtsystem.auditor","model_document_page","mgmtsystem.group_mgmtsystem_auditor",1,1,1,0
+"access_document_page_history_mgmtsystem_auditor","document.page.history.mgmtsystem.auditor","document_page.model_document_page_history","mgmtsystem.group_mgmtsystem_auditor",1,1,1,0
 "access_mgmtsystem_system_manager","mgmtsystem.system.manager","model_mgmtsystem_system","mgmtsystem.group_mgmtsystem_manager",1,1,1,1
+"access_document_page_mgmtsystem_manager","document.page.mgmtsystem.manager","model_document_page","mgmtsystem.group_mgmtsystem_manager",1,1,1,1
+
+
+
```
yajo: :fire:

mgmtsystem/security/mgmtsystem_security.xml
```diff
@@ -52,4 +52,89 @@
         <field name="domain_force">['|',('company_id','=',False),('company_id','child_of',[user.company_id.id])]</field>
     </record>
 
+    <!-- Read/Write/Create/Unlink access in 'Management System / viewer' group
+        only for 'Management system' documents -->
+    <record id="document_page_rule_mgmtsystem_viewer" model="ir.rule">
+        <field name="name">Document page for viewers</field>
+        <field name="model_id" ref="document_page.model_document_page"/>
+        <field name="domain_force">['|', ('management_system', '=', True), ('parent_id.management_system', '=', True)]</field>
```
yajo: What about a page marked as management_system=True that has a child page that has another child page? The 2nd child would be hidden? :thinking: 

Maybe we can use `child_of` somehow? Or then maybe it's better to remove `('parent_id.management_system', '=', True)`.

yajo (commented):
In general terms, in odoo users always have access to documents where they are followers. Do you think that approach fits fine here?

Shouldn't be a context `{'default_management_system': True}` in management system views then?

max3903:
@ernestotejeda If we want to achieve the feature you described, a one line fix in the access right file is enough. Or maybe I missed something.

I don't see the purpose for a new `management_system` field.

Regarding the record rules, the idea is to allow any employee to make changes to a document page and have those changes reviewed by the Approving group. We don't want a group to only read and another one to update. For example, anyone who can read should be able to fix a typo.

ernestotejeda:
@Yajo and @max3903, maybe I didn't explain very well the purpose of this module. The goal is that a management system user to have access only to the management system document.page. We don't want that the user belong to a knowledge access group, because then he/she would have access to all existing document.page (including those that are not of management system).

Apparently the use of the management_system field is not very well understood. The strategy is to set it to true only in the document.page with type=category created by the management system to find all document.page of management system.

We don't have to take into account the children of the children because currently that is not contemplated. Not even for an administrator user.

@max3903, I'm giving the possibility of having several levels of access on document.page of management system as same exists in the access groups of Knowledge. So we can have the user *management system / viewer* as a user that with read-only permission to everything, including the management system document.page.

max3903:
@ernestotejeda I see your point now. May I suggest the following approach:

1. a selection field `scope` on the document.page in the document_page module.
2. the module mgmtsystem_manual adds a `Management System` value in the previous selection field. It should also provide the access right and the record rule using the scope.

This way, other modules can add their own scope to the list instead of a new boolean field.

yajo:
Also I think that if that's the purpose, you should handle 2 cases:

1. Preexisting pages created for management systems should get that field properly filled.
1. Management system views should have a context such as `{'default_scope': 'management_system'}` so that users get that field filled by default when creating a page from there.

ernestotejeda:
I'm going to explain in a little more detail, because I have made a change to adjust it more according to what is needed, it is no longer necessary to add the Boolean field, and therefore the selection is not required, which seems to me anyway a good idea in terms of achieving genericity for other objectives, but the following happens here:

There are several modules of management system that create sub-menus as children of the Documentation menu to handle document.page related to management system.

The idea now is that in each of these modules, the permissions of the access groups of Management system to document.page are activated and a rule is created that restricts those permissions only to the document.page that will be shown in the corresponding submenu of the module. That is, the domain_force of the rule is going to be almost the same as the domain of the action associated with the menu. So each module that is created to add a sub-menu within the documentation menu will have to do the same.

I can not make a single general rule, because each module has its specificities. For example:

The module document_page_work_instruction adds the sub-menu *Work Instructions* inside the menu *Management system > Documentation* and it is associated with an action that has the following domain:
```eval = "[('parent_id', '=', ref ('document_page_group_work_instructions'))]]"```
therefore, that module is the one that should be responsible for putting a rule with the following domain_force:
```eval = "['|', ('parent_id', '=', ref ('document_page_group_work_instructions')), ('id', '=', ref ('document_page_group_work_instructions'))]]"```

But nevertheless:
The module mgmtsystem_manual adds the sub-menu *Manuals* inside the menu *Management system > Documentation* and it is associated with an action that has the following domain:
```eval = "[('parent_id', 'ilike', 'manual')]"```
therefore, that module is the one that should be responsible for putting a rule with the following domain_force:
```eval = "['|', ('parent_id', 'ilike', 'manual'), ('name', 'ilike', 'manual')]"```

File: document_page_environmental_aspect/__manifest__.py [hunk 3]
```diff
@@ -32,6 +32,7 @@
     "data": [
         'data/document_page.xml',
         'views/document_page.xml',
+        'security/environmental_aspect_security.xml',
     ],
     'installable': False,
```
martintorresvm: @ernestotejeda This module is `installable` False so not sure if I am missing something but any changes aren't notice for Travis for instance

martintorresvm (changes requested):

ernestotejeda:
@victormmtorres Changes done

martintorresvm (approved):

document_page_environmental_aspect/security/environmental_aspect_security.xml
```diff
@@ -0,0 +1,45 @@
...
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
+    </record>
+
+    <record id="environmental_document_history_rule_mgmtsystem_viewer" model="ir.rule">
+        <field name="name">Environmental document page for viewers</field>
+        <field name="model_id" ref="document_page.model_document_page_history"/>
+        <field name="domain_force" eval="['|', ('page_id.parent_id','=',ref('document_page_group_environmental_aspect')), ('page_id.id','=',ref('document_page_group_environmental_aspect'))]"/>
```
yajo:
```suggestion
        <field name="domain_force" eval="['|', ('page_id.parent_id','=',ref('document_page_group_environmental_aspect')), ('page_id','=',ref('document_page_group_environmental_aspect'))]"/>
```

mgmtsystem/security/ir.model.access.csv
```diff
@@ -1,5 +1,10 @@
-"id","name","model_id:id","group_id:id","perm_read","perm_write","perm_create","perm_unlink"
-"access_mgmtsystem_system_user","mgmtsystem.system.user","model_mgmtsystem_system","mgmtsystem.group_mgmtsystem_user",1,0,0,0
-"access_mgmtsystem_system_viewer","mgmtsystem.system.viewer","model_mgmtsystem_system","mgmtsystem.group_mgmtsystem_viewer",1,0,0,0
-"access_mgmtsystem_system_auditor","mgmtsystem.system.auditor","model_mgmtsystem_system","mgmtsystem.group_mgmtsystem_auditor",1,0,0,0
-"access_mgmtsystem_system_manager","mgmtsystem.system.manager","model_mgmtsystem_system","mgmtsystem.group_mgmtsystem_manager",1,1,1,1
```
yajo, line 5: Weird, where does this diff come from? Did you change the EOL character or something like that? :thinking:

mgmtsystem/security/ir.model.access.csv
```diff
@@ -1,5 +1,10 @@
-"id","name","model_id:id","group_id:id","perm_read","perm_write","perm_create","perm_unlink"
-"access_mgmtsystem_system_user","mgmtsystem.system.user","model_mgmtsystem_system","mgmtsystem.group_mgmtsystem_user",1,0,0,0
-"access_mgmtsystem_system_viewer","mgmtsystem.system.viewer","model_mgmtsystem_system","mgmtsystem.group_mgmtsystem_viewer",1,0,0,0
-"access_mgmtsystem_system_auditor","mgmtsystem.system.auditor","model_mgmtsystem_system","mgmtsystem.group_mgmtsystem_auditor",1,0,0,0
-"access_mgmtsystem_system_manager","mgmtsystem.system.manager","model_mgmtsystem_system","mgmtsystem.group_mgmtsystem_manager",1,1,1,1
+"id","name","model_id:id","group_id:id","perm_read","perm_write","perm_create","perm_unlink"
+"access_mgmtsystem_system_user","mgmtsystem.system.user","model_mgmtsystem_system","mgmtsystem.group_mgmtsystem_user",1,0,0,0
+"access_mgmtsystem_system_viewer","mgmtsystem.system.viewer","model_mgmtsystem_system","mgmtsystem.group_mgmtsystem_viewer",1,0,0,0
+"access_document_page_mgmtsystem_viewer","document.page.mgmtsystem.viewer","document_page.model_document_page","mgmtsystem.group_mgmtsystem_viewer",1,0,0,0
```
yajo:
```suggestion
"access_document_page_mgmtsystem_viewer","document.page.mgmtsystem.viewer","document_page.model_document_page","group_mgmtsystem_viewer",1,0,0,0
```

mgmtsystem/security/ir.model.access.csv
```diff
@@ -1,5 +1,10 @@
...
+"access_document_page_history_mgmtsystem_viewer","document.page.history.mgmtsystem.viewer","document_page.model_document_page_history","mgmtsystem.group_mgmtsystem_viewer",1,0,0,0
```
yajo:
```suggestion
"access_document_page_history_mgmtsystem_viewer","document.page.history.mgmtsystem.viewer","document_page.model_document_page_history","group_mgmtsystem_viewer",1,0,0,0
```

mgmtsystem/security/ir.model.access.csv
```diff
@@ -1,5 +1,10 @@
...
-"access_mgmtsystem_system_manager","mgmtsystem.system.manager","model_mgmtsystem_system","mgmtsystem.group_mgmtsystem_manager",1,1,1,1
+"id","name","model_id:id","group_id:id","perm_read","perm_write","perm_create","perm_unlink"
+"access_mgmtsystem_system_user","mgmtsystem.system.user","model_mgmtsystem_system","mgmtsystem.group_mgmtsystem_user",1,0,0,0
+"access_mgmtsystem_system_viewer","mgmtsystem.system.viewer","model_mgmtsystem_system","mgmtsystem.group_mgmtsystem_viewer",1,0,0,0
+"access_document_page_mgmtsystem_viewer","document.page.mgmtsystem.viewer","document_page.model_document_page","mgmtsystem.group_mgmtsystem_viewer",1,0,0,0
+"access_document_page_history_mgmtsystem_viewer","document.page.history.mgmtsystem.viewer","document_page.model_document_page_history","mgmtsystem.group_mgmtsystem_viewer",1,0,0,0
+"access_mgmtsystem_system_auditor","mgmtsystem.system.auditor","model_mgmtsystem_system","mgmtsystem.group_mgmtsystem_auditor",1,0,0,0
+"access_document_page_mgmtsystem_auditor","document.page.mgmtsystem.auditor","document_page.model_document_page","mgmtsystem.group_mgmtsystem_auditor",1,1,1,0
```
yajo:
```suggestion
"access_document_page_mgmtsystem_auditor","document.page.mgmtsystem.auditor","document_page.model_document_page","group_mgmtsystem_auditor",1,1,1,0
```

mgmtsystem/security/ir.model.access.csv
```diff
@@ -1,5 +1,10 @@
...
+"access_document_page_history_mgmtsystem_auditor","document.page.history.mgmtsystem.auditor","document_page.model_document_page_history","mgmtsystem.group_mgmtsystem_auditor",1,1,1,0
```
yajo:
```suggestion
"access_document_page_history_mgmtsystem_auditor","document.page.history.mgmtsystem.auditor","document_page.model_document_page_history","group_mgmtsystem_auditor",1,1,1,0
```

mgmtsystem/security/ir.model.access.csv
```diff
@@ -1,5 +1,10 @@
...
+"access_mgmtsystem_system_manager","mgmtsystem.system.manager","model_mgmtsystem_system","mgmtsystem.group_mgmtsystem_manager",1,1,1,1
+"access_document_page_mgmtsystem_manager","document.page.mgmtsystem.manager","document_page.model_document_page","mgmtsystem.group_mgmtsystem_manager",1,1,1,1
```
yajo:
```suggestion
"access_document_page_mgmtsystem_manager","document.page.mgmtsystem.manager","document_page.model_document_page","group_mgmtsystem_manager",1,1,1,1
```

mgmtsystem/security/mgmtsystem_security.xml
```diff
@@ -52,4 +52,94 @@
         <field name="domain_force">['|',('company_id','=',False),('company_id','child_of',[user.company_id.id])]</field>
     </record>
 
+    <!-- =================== -->
```
yajo:
```suggestion
    <!-- No page access by default; submodules add access to specific sections -->
```

mgmtsystem/security/mgmtsystem_security.xml
```diff
@@ -52,4 +52,94 @@
...
+        <field name="name">No document page history for viewers</field>
+        <field name="model_id" ref="document_page.model_document_page_history"/>
+        <field name="domain_force" eval="[(0, '=', 1)]"/>
+        <field name="groups" eval="[(4, ref('mgmtsystem.group_mgmtsystem_viewer'))]"/>
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
+    </record>
+    <!-- =================== -->
```
yajo: These kind of comment lines are unnecessary. As any other piece of code, it's a good practice to remove the unnecessary stuff. :wink: 

```suggestion
```

mgmtsystem/security/mgmtsystem_security.xml
```diff
@@ -52,4 +52,94 @@
...
+
+    <!-- =================== -->
```
yajo:
```suggestion
```

mgmtsystem/security/mgmtsystem_security.xml
```diff
@@ -52,4 +52,94 @@
...
+    <!-- Read access in 'Knowledge / Knowledge User' group for all documents -->
+    <record id="document_page_rule_document_user_r" model="ir.rule">
+        <field name="name">All document pages (read)</field>
+        <field name="model_id" ref="document_page.model_document_page"/>
+        <field name="domain_force">[(1, '=', 1)]</field>
+        <field name="groups" eval="[(4, ref('knowledge.group_document_user'))]"/>
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="False"/>
```
yajo: You can set to `True` safely because those will only be true if the relevant `ir.model.access` is also true.

```suggestion
        <field name="perm_write" eval="True"/>
```

mgmtsystem/security/mgmtsystem_security.xml
```diff
@@ -52,4 +52,94 @@
...
+        <field name="perm_create" eval="False"/>
```
yajo:
```suggestion
        <field name="perm_create" eval="True"/>
```

mgmtsystem/security/mgmtsystem_security.xml
```diff
@@ -52,4 +52,94 @@
...
+        <field name="perm_unlink" eval="False"/>
```
yajo:
```suggestion
        <field name="perm_unlink" eval="True"/>
```

mgmtsystem/security/mgmtsystem_security.xml
```diff
@@ -52,4 +52,94 @@
...
+        <field name="active" eval="False"/>
```
yajo: Why is this disabled? :thinking: 

```suggestion
```

mgmtsystem/security/mgmtsystem_security.xml
```diff
@@ -52,4 +52,94 @@
...
+    </record>
+
+    <record id="document_page_history_rule_document_user_r" model="ir.rule">
+        <field name="name">All document pages history (read)</field>
+        <field name="model_id" ref="document_page.model_document_page_history"/>
+        <field name="domain_force">[(1, '=', 1)]</field>
+        <field name="groups" eval="[(4, ref('knowledge.group_document_user'))]"/>
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="False"/>
```
yajo:
```suggestion
        <field name="perm_write" eval="True"/>
```

mgmtsystem/security/mgmtsystem_security.xml
```diff
@@ -52,4 +52,94 @@
...
+        <field name="perm_create" eval="False"/>
```
yajo:
```suggestion
        <field name="perm_create" eval="True"/>
```

mgmtsystem/security/mgmtsystem_security.xml
```diff
@@ -52,4 +52,94 @@
...
+        <field name="perm_unlink" eval="False"/>
```
yajo:
```suggestion
        <field name="perm_unlink" eval="True"/>
```

mgmtsystem/security/mgmtsystem_security.xml
```diff
@@ -52,4 +52,94 @@
...
+        <field name="active" eval="False"/>
```
yajo:
```suggestion
```

mgmtsystem/security/mgmtsystem_security.xml
```diff
@@ -52,4 +52,94 @@
...
+    </record>
+    <!-- =================== -->
```
yajo:
```suggestion
```

mgmtsystem/security/mgmtsystem_security.xml
```diff
@@ -52,4 +52,94 @@
...
+
+    <!-- Read/Write/Create access in 'Knowledge / Editor' group for all documents -->
```
yajo: All this code from here to the bottom is probably unneeded if you apply above suggestions, because groups `document_page.group_document_editor` and `document_page.group_document_manager` [inherit from `knowledge.group_document_user`](https://github.com/OCA/knowledge/blob/f13cfb32ef24c88a6e86d65cd2e788f8b52f89e9/document_page/security/document_page_security.xml#L7).

The `ir.rule` above will grant all kinds of access, and then it will be the `ir.model.access` record who decides if the user really has it or not.

mgmtsystem_manual/security/manual_security.xml
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <!-- Read/Write/Create/Unlink access in 'Management System / viewer' group
+        to 'Manual' documents -->
+    <record id="manual_document_rule_mgmtsystem_viewer" model="ir.rule">
+        <field name="name">Manual document page for viewers</field>
+        <field name="model_id" ref="document_page.model_document_page"/>
+        <field name="domain_force" eval="['|', ('parent_id','ilike', 'manual'), ('name','ilike', 'manual')]"/>
```
yajo, line 8: This kind of "security" where containing the word "manual" in the name grants/revokes access is ridiculous :roll_eyes:. Please take the chance to change this.

```suggestion
        <field name="domain_force" eval="['|', ('parent_id','=', ref('manuals')), ('id','=', ref('manuals'))]"/>
```

Remember to change it here too:
https://github.com/OCA/management-system/blob/91b17bd4af2d8ae90fb0d6cfb95b60f417d5942c/mgmtsystem_manual/views/mgmtsystem_manual.xml#L32

mgmtsystem_manual/security/manual_security.xml
```diff
@@ -0,0 +1,45 @@
...
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
+    </record>
+
+    <record id="manual_document_history_rule_mgmtsystem_viewer" model="ir.rule">
+        <field name="name">Manual document page for viewers</field>
+        <field name="model_id" ref="document_page.model_document_page_history"/>
+        <field name="domain_force" eval="['|', ('page_id.parent_id','ilike', 'manual'), ('page_id.name','ilike', 'manual')]"/>
```
yajo, line 19:
```suggestion
        <field name="domain_force" eval="['|', ('page_id.parent_id','=', ref('manuals')), ('page_id','=', ref('manuals'))]"/>
```

mgmtsystem_manual/security/manual_security.xml
```diff
@@ -0,0 +1,45 @@
...
+        <field name="groups" eval="[(4, ref('mgmtsystem.group_mgmtsystem_viewer'))]"/>
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
+    </record>
+    <!-- =================== -->
```
yajo: Delete from this line to the end.

document_page_work_instruction/security/work_instruction_security.xml
```diff
@@ -0,0 +1,45 @@
...
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
+    </record>
+
+    <record id="work_document_history_rule_mgmtsystem_viewer" model="ir.rule">
+        <field name="name">Work document page history for viewers</field>
+        <field name="model_id" ref="document_page.model_document_page_history"/>
+        <field name="domain_force" eval="['|', ('page_id.parent_id','=',ref('document_page_group_work_instructions')), ('page_id.id','=',ref('document_page_group_work_instructions'))]"/>
```
yajo:
```suggestion
        <field name="domain_force" eval="['|', ('page_id.parent_id','=',ref('document_page_group_work_instructions')), ('page_id','=',ref('document_page_group_work_instructions'))]"/>
```

document_page_work_instruction/security/work_instruction_security.xml
```diff
@@ -0,0 +1,45 @@
...
+        <field name="groups" eval="[(4, ref('mgmtsystem.group_mgmtsystem_viewer'))]"/>
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
+    </record>
+    <!-- =================== -->
```
yajo: :fire: Delete from here to the end.

document_page_procedure/security/procedure_security.xml
```diff
@@ -0,0 +1,45 @@
...
+        <field name="name">Procedure document page for viewers</field>
+        <field name="model_id" ref="document_page.model_document_page_history"/>
+        <field name="domain_force" eval="['|', ('page_id.parent_id','=',ref('document_page_group_procedure')), ('page_id.id','=',ref('document_page_group_procedure'))]"/>
+        <field name="groups" eval="[(4, ref('mgmtsystem.group_mgmtsystem_viewer'))]"/>
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
+    </record>
+    <!-- =================== -->
```
yajo: :fire: Delete from here to the end.

document_page_procedure/security/procedure_security.xml
```diff
@@ -0,0 +1,45 @@
...
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
+    </record>
+
+    <record id="procedure_document_history_rule_mgmtsystem_viewer" model="ir.rule">
+        <field name="name">Procedure document page for viewers</field>
+        <field name="model_id" ref="document_page.model_document_page_history"/>
+        <field name="domain_force" eval="['|', ('page_id.parent_id','=',ref('document_page_group_procedure')), ('page_id.id','=',ref('document_page_group_procedure'))]"/>
```
yajo:
```suggestion
        <field name="domain_force" eval="['|', ('page_id.parent_id','=',ref('document_page_group_procedure')), ('page_id','=',ref('document_page_group_procedure'))]"/>
```

document_page_environmental_aspect/security/environmental_aspect_security.xml
```diff
@@ -0,0 +1,45 @@
...
+        <field name="groups" eval="[(4, ref('mgmtsystem.group_mgmtsystem_viewer'))]"/>
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
+    </record>
+    <!-- =================== -->
```
yajo: Delete from here to the end.

yajo (changes requested):
Quite a bit of pending changes, as we discussed some time ago. Just publishing them here as a reminder. :blush: 

Also I noticed that there's a group that it seems like is doing nothing: `mgmtsystem.group_mgmtsystem_user_manager`. Could we take this chance to remove it?:

https://github.com/OCA/management-system/blob/91b17bd4af2d8ae90fb0d6cfb95b60f417d5942c/mgmtsystem/security/mgmtsystem_security.xml#L23-L30

mgmtsystem/security/mgmtsystem_security.xml
ernestotejeda: I did this thinking that if someone wanted to install only the mgmtsystem module and not document_page_work_instruction, document_page_procedure or any other that added a registration rule related to document page, it would not make sense to apply this rule, but on second thought, it doesn't make much sense that you only install this module, because it is a base module. So, i'm going to commit this suggestion.

pedrobaeza:
@ernestotejeda please avoid the GitHub edition tool that creates a commit for each change.

mgmtsystem/security/ir.model.access.csv
ernestotejeda: I see something weird too related with EOL but I don't know what it is.

mgmtsystem_manual/security/manual_security.xml
ernestotejeda: I agree with you that this is wrong, but I am not sure that the way to solve it is the one you propose. I think we have to think a little more about some solution for this.

mgmtsystem_manual/security/manual_security.xml
ernestotejeda: The same in here

ernestotejeda:
I accepted and applied all the changes that you suggested, except those where I have answered the comment:

https://github.com/OCA/management-system/pull/241#discussion_r273844211

https://github.com/OCA/management-system/pull/241#discussion_r273981958

https://github.com/OCA/management-system/pull/241#discussion_r273982202

mgmtsystem/security/ir.model.access.csv
yajo: Usually windows puts a CRLF end, and Linux a LF end. If your editor is programmed to fix EOLs and the original author developed this with CRLF, it can be a problem.

Well, I used github's feature to ignore whitespace changes to see this diff, so don't worry too much about it.

mgmtsystem_manual/security/manual_security.xml
yajo: What else do you want to think? :thinking: I just suggest you to let it as it is in all other addons.

Doing it this way has 2 problems:

1. If a user names a top page as *manual* whatever, unexpected users will have access to it.
2. It doesn't take translations into account.

I'm not saying my solution is perfect, but at least it should be better than it is right now...

pedrobaeza:
@Yajo @ernestotejeda please both synchronize for finishing this

rafaelbn:
@ernestotejeda @Yajo as @pedrobaeza said, synchronize for finishing this PR thanks

rafaelbn (approved):
:+1:

pedrobaeza:
/ocabot merge minor

@ernestotejeda please migrate it to v12

PR #255, opened 2019-07-11
Add default menu to main categories

simahawk:
Follow-up of https://github.com/OCA/knowledge/pull/218

### DIFF
--- a/mgmtsystem_action/tests/test_create_action.py
+++ b/mgmtsystem_action/tests/test_create_action.py
@@ -1,15 +1,9 @@
 
 from odoo import exceptions
 from odoo.tests import common
-from datetime import datetime, timedelta
+from datetime import datetime
+from freezegun import freeze_time
 import mock
-import time
-
-
-def freeze_time(dt):
-    mock_time = mock.Mock()
-    mock_time.return_value = time.mktime(dt.timetuple())
-    return mock_time
 
 
 class TestModelAction(common.SavepointCase):
@@ -70,18 +64,17 @@ class TestModelAction(common.SavepointCase):
             '&id={}&model={}'.format(self.record.id, self.record._name), url
         )
 
+    @freeze_time('2019-06-05')
     def test_process_reminder_queue(self):
         """Check if process_reminder_queue work when days reminder are 10."""
-        ten_days_date = datetime.now().date() + timedelta(days=10)
         self.record.write({
-            'date_deadline': ten_days_date,  # 10 days from now
+            'date_deadline': '2019-06-15',  # 10 days from now
             'stage_id': self.env.ref('mgmtsystem_action.stage_open').id,
         })
-        with mock.patch('time.time', freeze_time(ten_days_date)):
-            tmpl_model = self.env['mail.template']
-            with mock.patch.object(type(tmpl_model), 'send_mail') as mocked:
-                self.env['mgmtsystem.action'].process_reminder_queue()
-                mocked.assert_called_with(self.record.id)
+        tmpl_model = self.env['mail.template']
+        with mock.patch.object(type(tmpl_model), 'send_mail') as mocked:
+            self.env['mgmtsystem.action'].process_reminder_queue()
+            mocked.assert_called_with(self.record.id)
 
     def test_stage_groups(self):
         """Check if stage_groups return all stages."""
--- a/mgmtsystem_action/tests/test_create_action.py
+++ b/mgmtsystem_action/tests/test_create_action.py
@@ -1,9 +1,14 @@
 
 from odoo import exceptions
 from odoo.tests import common
-from datetime import datetime
-from freezegun import freeze_time
 import mock
+import time
+
+
+def freeze_time(dt):
+    mock_time = mock.Mock()
+    mock_time.return_value = time.mktime(dt.timetuple())
+    return mock_time
 
 
 class TestModelAction(common.SavepointCase):
@@ -64,17 +69,18 @@ class TestModelAction(common.SavepointCase):
             '&id={}&model={}'.format(self.record.id, self.record._name), url
         )
 
-    @freeze_time('2019-06-05')
     def test_process_reminder_queue(self):
         """Check if process_reminder_queue work when days reminder are 10."""
+        ten_days_date = datetime.now().date() + timedelta(days=10)
         self.record.write({
-            'date_deadline': '2019-06-15',  # 10 days from now
+            'date_deadline': ten_days_date,  # 10 days from now
             'stage_id': self.env.ref('mgmtsystem_action.stage_open').id,
         })
-        tmpl_model = self.env['mail.template']
-        with mock.patch.object(type(tmpl_model), 'send_mail') as mocked:
-            self.env['mgmtsystem.action'].process_reminder_queue()
-            mocked.assert_called_with(self.record.id)
+        with mock.patch('time.time', freeze_time(ten_days_date)):
+            tmpl_model = self.env['mail.template']
+            with mock.patch.object(type(tmpl_model), 'send_mail') as mocked:
+                self.env['mgmtsystem.action'].process_reminder_queue()
+                mocked.assert_called_with(self.record.id)        
 
     def test_stage_groups(self):
         """Check if stage_groups return all stages."""
--- a/mgmtsystem_action/tests/test_create_action.py
+++ b/mgmtsystem_action/tests/test_create_action.py
@@ -1,6 +1,7 @@
 
 from odoo import exceptions
 from odoo.tests import common
+from datetime import datetime, timedelta
 import mock
 import time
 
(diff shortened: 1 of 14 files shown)


i-vyshnevska (commented):
in manifest need to change sequence of loading data

simahawk:
yep, I'm on it ;)

simahawk:
damn, there are inter-relations, I have to think about something better than reordering

i-vyshnevska:
one more record after menuitem to add menu_id

simahawk:
splitted menu definition. Should work now.

simahawk:
there's a test failing on mgmtsystem_action unrelated to this change.
12.0 is red as well. I'll check.

simahawk:
:tada: :green_heart:

simahawk:
@i-vyshnevska @max3903 can you drop a review? This fixes 12.0 branch as well.

i-vyshnevska (approved):

vrenaville (approved):

pedrobaeza:
Doesn't this require migration scripts?

simahawk:
in theory just an update of the modules.

pedrobaeza:
But I don't get why are you now just defining the action of the menu instead of full menu definition.

simahawk:
chicken-n-egg problem: we need the page to link the menu but the menu needs the action which in turns need the page record for domain and ctx

pedrobaeza:
But the menus fields are completed when?

simahawk:
what do you mean w/ "completed"? Populated?
As I reported on OCA/knowledge#218 the menu field is not editable UNLESS you use the action "Create menu" which is going to populate it.
So, by default, thanks to the change here, the URL is going to be built using the menu item/action of the main category.
If you set a custom menu, is going to use that.

pedrobaeza:
Yes, I mean populated. My fear is that someone installing the modules won't have the menus in their places.

simahawk:
how? The menu ARE  in the modules and they were already there, I did not add them... I  don't get your point :man_shrugging:

simahawk:
you mean if deleted maybe?
I that case if you upgrade the module they gonna be re-created as they are not noupdate AFAIR

pedrobaeza:
But you are removing their parents and sequences...

marcelsavegnago (approved):

simahawk:
> But you are removing their parents and sequences...

please check better, I'm not removing anything. They are exactly the same + the menu reference.
I split their creation and the setting of the action because we need them to be there when the page is created but we don't have the action yet.

simahawk:
rebased

mgmtsystem_action/tests/test_create_action.py
```diff
@@ -70,18 +64,17 @@ def test_get_action_url(self):
             '&id={}&model={}'.format(self.record.id, self.record._name), url
         )
 
+    @freeze_time('2019-06-05')
```
pedrobaeza: Please don't introduce libraries like this one only for testing and more without a strong need.

pedrobaeza (changes requested):

mgmtsystem_action/tests/test_create_action.py
simahawk: it is a strong need because is the only way to test the real feature of sending emails in the future as it would work out of tests.

Also, this lib is used already in other OCA repos https://github.com/search?q=org%3AOCA+freezegun&type=Code

Many times I see hacks to test such cases (like dates passed in ctx only for testing) and that's worse than adding an external lib "just" for testing IMO.

pedrobaeza: It has been used only in your devs, not in whole OCA.

Odoo has included a tool for doing that in tests, but not sure if it's for v13 or v12. Any way, I don't want this hidden dependency breaking my internal CIs as I'm using these modules. The test was previously there and it was working, and it's totally unrelated to the purpose of this PR.

simahawk: I  know the test was there, I coded it ;)
"my devs"? If you mean @guewen et al, yes... :)
BUT it was failing on this new PR and it was a custom hack I've decided to use a well know library to handle it.
Is not hidden, it's declared in requirements.txt and I bet you check that or include that when you add a new repository to your project. 
If not, then I  wonder how you pick dependencies for new repos :thinking: 
Regarding the v12/13 feature: I don't see anything. Any pointer?

pedrobaeza: We don't aggregate all the requirements.txt of all the used repositories because we don't use every module!

I tweeted some time ago about it, but I'm not able to find it right now, and Twitter search is terrible (well, it doesn't exists at all).

Do the import/test conditionally at least for not being executed if no library present.

guewen: > We don't aggregate all the requirements.txt of all the used repositories because we don't use every module!

Maybe that's what we should fix?

Reducing the test coverage is probably not the best course of action when the problem is only a missing pip install?

pedrobaeza: I don't want extra burden Python libraries in my image if not used, as there are also other things to perform with some pip installs (for example, requiring packages for sources, or other libraries, like mysql and similar in server-backend repo).

Any way, the main problem here is changing this on an already stable module. This will break my current instances CI with this module deployed.

BT-cserra: Sorry to get in the middle, but I came across this one because I also use a lot ```freezegun``` to make tests that traverse in time.

@simahawk : If for whathever reason ```freezegun``` is banned on OCA and you still want to contribute this, you can use ```mock``` instead, which should not be a problem because it is listed as part of Odoo's ```requirements.txt```. For instance, Odoo's core (V12) uses it on this test to go back in time: https://github.com/odoo/odoo/blob/e1833936255839cd34eba5758ca4c3124fd2522c/addons/account/tests/test_account_customer_invoice.py#L394. However in my tests I had to patch also ```Datetime.now``` because sometimes internally Odoo uses a datetime, this way:

```python
from mock import patch
# ...
# Then inside the test:
with patch('odoo.fields.Date.today', lambda: datetime.date(2018, 1, 3)),
        patch('odoo.fields.Datetime.now', lambda: datetime.datetime(2018, 1, 3)):
    # The code that tests in the past.
    # ...
```

It's more painful to use than ```freezegun```, but can be an alternative to it.

simahawk: @BT-cserra tnx for the feedback. Yes, I did that sometimes but reality is the - as you said - is a bit painful because it's not granted to be applied to every places whereas a date or time is implied. 
On the contrary, freezegun does exactly what you expect, no matter the place.

Regarding the ban, frankly I don't understand @pedrobaeza ... to me is really not a point.
To me this lib should stay everywhere in OCA and ppl should be encouraged to use it instead of mocking things every time in a different way.

pedrobaeza: I don't ban the library itself. What I ban is to introduce it in a module already landed in stable, making existing deployments with linked CIs to fail.

simahawk: then I wonder how do you update your stable project's CI?
You don't check what's coming thru? You don't take care of checking what's changed?
And then, even if it breaks your CI, is going to be an easy fix: add the lib to requirements and you are done.
My friend, honestly I think we are just losing time here, sorry :angel:

pedrobaeza: Of course it will fail, but I don't want to change all my scaffolding for introducing that library in them. My reasoning is that you are using it only marginally here, so a workaround like the propose one is good for that. If you would introduce it at the beginning, then it's part of the deployment you need to do when the module is introduced in the instance.

pedrobaeza: And I repeat that the title and the purpose of this PR has nothing to do with this change.

guewen: > Of course it will fail, but I don't want to change all my scaffolding for introducing that library in them. My reasoning is that you are using it only marginally here, so a workaround like the propose one is good for that. If you would introduce it at the beginning, then it's part of the deployment you need to do when the module is introduced in the instance.

So then we have to use workarounds everywhere because you cannot update your scaffolding? Doesn't really make sense to me.

guewen: > making existing deployments with linked CIs to fail.

Isn't it the purpose of the CI by the way, so you can, for instance detect and add the missing libs? :wink:

pedrobaeza: @guewen there are 3 main reasons for not liking this:

- The PR was not for that. See the title and the purpose of it.
- It's introducing the dependency on a stable branch well established, with several customers that already have this module deployed.
- This is something only for testing, and only in one place, not adding any value to the module itself.

dreispt (approved):

PR #256, opened 2019-07-24
[FIX] mgmtsystem_action: change static date value for current date

ernestotejeda:
Cc @tecnativa


max3903 (approved):

mgmtsystem_action/__manifest__.py
```diff
@@ -22,7 +22,6 @@
         'security/mgmtsystem_action_security.xml',
         'data/action_sequence.xml',
         'views/mgmtsystem_action.xml',
-        'views/mgmtsystem_action_kanban.xml',
```
pedrobaeza, line 25: Why are you removing this kanban?

ernestotejeda: The only thing we have in this file is the same view kanban view of mgmtsystem.action that is in the mgmtsystem_action.xml file.

https://github.com/OCA/management-system/blob/e3704f65a2126b08023221b4180c3c26d2dde1b0/mgmtsystem_action/views/mgmtsystem_action_kanban.xml#L3

https://github.com/OCA/management-system/blob/e3704f65a2126b08023221b4180c3c26d2dde1b0/mgmtsystem_action/views/mgmtsystem_action.xml#L122

They are exactly the same. It was one of the pylint errors when I pushed the branch for the first time

pedrobaeza: OK, thanks for the explanation

pedrobaeza (approved):
I usually prefer static dates against dynamic ones, as we later faced problems with leap years, different weeks, etc, but I don't think in this case it's a problem.

PR #270, opened 2019-11-08
13.0 mig mgmtsystem survey

ernestotejeda:
Depends on:
- [x] https://github.com/OCA/management-system/pull/268

~~As this module depends on https://github.com/OCA/management-system/pull/268 I have included here the commits of that PR so it can be functionally tested.~~

There is no survey.stage model in v13, now the survey.survey model has a selection field instead of having a many2one relationship with survey.stage. So I adapted the module according to those changes

Cc @Tecnativa TT20485


ernestotejeda:
As https://github.com/OCA/management-system/pull/268 is merged i rebased and cleaned up the commit history a bit

Tardo (approved):
:+1: LGTM

mgmtsystem_survey/models/survey_survey.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2019 Tecnativa - Ernesto Tejeda
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class Survey(models.Model):
+    _inherit = "survey.survey"
+
+    state = fields.Selection(selection_add=[("management_system", "Management System")])
```
pedrobaeza, line 10: You must create a pre-migration script for this that sets this value in if previous stage was `mgmtsystem_survey_type`

pedrobaeza (changes requested):

ernestotejeda:
changes done

mgmtsystem_survey/migrations/13.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,18 @@
...
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE survey_survey s
+        SET state = 'management_system',
+        FROM survey_stage ss
+        WHERE s.stage_id = ss.id
```
pedrobaeza: Don't use this flaw connection. Instead get the `res_id` of the XML-ID:
```suggestion
        LEFT JOIN ir_model_data imd ON imd.res_id = s.stage_id AND imd.module = 'mgmt_survey' AND imd.name = 'mgmtsystem_survey_type'
```

ernestotejeda:
Changes done

pedrobaeza (approved):

PR #272, opened 2019-11-12
[13.0][MIG] document_page_work_instruction: Migration to 13.0

ernestotejeda:
Cc @Tecnativa TT20478


Tardo (approved):
:+1: LGTM

For me, the diff table can be improved using 100% of the width... 

** Update oca_dependencies when ready

ernestotejeda:
> For me, the diff table can be improved using 100% of the width...

I think it is a bit difficult to have the table width 100% and at the sam time to have the text wrapped.
What happens is that by default the python library that is used for the differences by default does not wrap the text and also the texts that are being taken to obtain the differences are html text and also those texts don't have spaces, then each paragraph (&lt;p&gt;) for example, although you see it with space, spaces are replaced by **&nbsp**; for the comparision. Maybe it needs more thinking and more time.

ernestotejeda:
oca_dependencies.txt file updated

pedrobaeza (approved):

PR #274, opened 2019-11-12
[13.0][MIG] document_page_quality_manual: Migration to 13.0

ernestotejeda:
Cc @Tecnativa TT20477


ernestotejeda:
I reduced the commit history a bit

ernestotejeda:
I have removed the **mgmtsystem_manual** dependency because this module does not use any features or alters any resources from **mgmtsystem_manual**. I don't know if there is any other reason I am missing that I need to maintain that dependence.
Technically it only depends on **document_page**
ping @pedrobaeza .

pedrobaeza:
If no technical dependency, then better to not force it.

Tardo (approved):
:+1: LGTM

** Update oca_dependencies when ready

ernestotejeda:
oca_dependencies.txt file updated

pedrobaeza (approved):

Issue #263, opened 2019-10-17
[RFC] Management System extension for production


mymage:
To continue the discussion started on the mailing list Contributors [https://odoo-community.org/groups/contributors-15/contributors-144648?mode=date&date_begin=2019-09-01&date_end=2019-10-01](https://odoo-community.org/groups/contributors-15/contributors-144648?mode=date&date_begin=2019-09-01&date_end=2019-10-01), we propose a module to add some features to fit some manufacturing needs.

Identification of the Department, the Workcenter and the Product of the NC, record of the quantity checked and not compliant, Partner reference for quality communications, Actions templates and efficacy indication.

You can find the module here 
[https://gitlab.com/PNLUG/Odoo/management-system-improvements/tree/11.0/mgmtsystem_extended](https://gitlab.com/PNLUG/Odoo/management-system-improvements/tree/11.0/mgmtsystem_extended)

These Issued is connected to the [OCA/manufacture#425](https://github.com/OCA/manufacture/issues/425) Issue for the Manufacture repository.

Please give us your opinion, our purpose is to incorporate it in the OCA repository.

At the moment the module is for the 11.0 version.

max3903:
Hello @mymage,

Thanks for the contributions and welcome to the OCA!

This looks like a big module with a lot of features and dependencies. To include it in the OCA and have a higher chance of adoption, we would need to make sure people can use only the features they need. We usually achieve this with one feature = one module, but we can also have one module with one setting per feature.

I don't know how much effort you want to put into this but here are couple options:
1. contribute your module as is and as a beta module. Make sure the CI (Travis and Runbot) are green
2. split your module and push one module at a time. As a beta (no unit tests) or stable (with unit tests) one.

You can also mix those options with the Odoo version, for example: use option 1 on version 11 and option 2 on version 12.

mymage:
@max3903 Thanks for the first suggestions. I think that for this module is better to ad the configuration option so I will work in that direction. In the mean time I will try it on runbot and for this I have to do a PR (right?): there are some specific things to follow to open a "test" PR?

ivantodorovich:
Hi @mymage !

This is very interesting :) Thanks !!

I think some of the features could be improvements done directly on the current modules.
For instance:

- Immediate Actions: replace the reference of a single Immediate action with one to multiple Immediate Actions

That's something I also thought about improving a while ago, and IMO this should be commited directly into mgmtsystem_nonconformity (with proper migration script!).

---

<dl>
[...]

PR #284, opened 2020-01-26
[11.0] [ADD] new module mgmtsystem_nonconformity_type

mymage:
Ref. #263

PR #286, opened 2020-02-02
[11.0] [ADD] new module mgmtsystem_partner

mymage:
Ref. #263


marcelofrare (approved):
It works properly for me

mymage:
I opened #348 for more audience :-)

marcelsavegnago:
> I opened #348 for more audience :-)

https://github.com/OCA/management-system/pull/348#issuecomment-657112219

>  I don't know what the effort would be, but it may be more interesting to reuse the contact register by creating a new type called Quality. What do you think ?
> 
>https://github.com/odoo/odoo/blob/39dbc8b0935a44c3457938076ff5688fc0800414/odoo/addons/base/models/res_partner.py#L180

image

mymage:
@etobella Can you kindly review this PR that locks #284, thanks

marcelsavegnago (approved):

etobella:
Just a small question, why not adding this change directly to managament system base module?

mymage:
> 
> 
> Just a small question, why not adding this change directly to managament system base module?

I'm not so confident with OCA world and the way things are managed. A suggestion that I got is to mantain separate modules for different funcionalities, but I have to learn where is the limit to ad a new one to an existing module or to do it all new.

So, in this case, thanks to your observation, I think it could be added to the base module*.
Do you think that is better to wait someone's other comment here or I can star a new PR do update another module?

*Thinking about that anyway, there is the problem that you can have different management systems, not only quality, with different contact reference so probabily it could be better to create a mgmtsystem_partner module where adding specific tab for "primary" systems (eg. quality, safety, ...).

etobella:
Well, if this is the case, maybe we should need to add the configuration for each managment system. For example:

- `mgmtsystem_quality` adds `quality` partner
- `mgmtsystem_safety` (it currently does not exist) adds `safety` partner

And so on.

In any case, I think we should need to wait for another comment from a PSC member @OCA/management-systems-maintainers

mymage:
@OCA/management-systems-maintainers
After a while I'm back here to try to close this PR part of a group. I know that it is for an old version but I will migrate it formward like other as soon as it will be approved.
Please let me know if there is something not right in the proposal.
Thanks

primes2h (approved):
LGTM

ivantodorovich (approved):

PR #315, opened 2020-03-23
[13.0][MIG] document_page_environmental_aspect

flachica:


flachica:
@ivantodorovich 

I don't know if the problem I'm having with squash is just mine or the branch has something wrong. 

The truth is that when I do 

`git format-patch --keep-subject --stdout origin/$(echo $to_version)..origin/$(echo $from_version) -- $(echo $module) | git am -3 --keep` as commented [here](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0)

git tells me that are having trouble applying the patches; that if I prefer to skip the steps. 

I have run `git am --skip` for every conflict. 
Then I have incorporated my changes and executed `git rebase -i origin/13.0` and the result is [this](https://github.com/OCA/management-system/pull/315).

I get lost. I'm trying to arm myself with patience.

Thank you for all

PR #323, opened 2020-04-02
[13.0][MIG] mgmtsystem_info_security_manual

flachica:
Migration to 13.0


dreispt (changes requested):
If you rename the module, we should provide the migration to allow people an easy upgrade.
Let me know if you prefer to keep the original name or need help on the mig script.

flachica:
> If you rename the module, we should provide the migration to allow people an easy upgrade.
> Let me know if you prefer to keep the original name or need help on the mig script.

I need help on the mig script. I think that the correct name is mgmtsystem_information_security. I prefer rename

FernandoRomera:
View a example in https://apps.odoo.com/apps/modules/12.0/web_pwa_oca/

In case you previously installed web_pwa, run the following steps with odoo shell, after having installed openupgradelib:

> from openupgradelib import openupgrade
> openupgrade.update_module_names(env.cr, [('web_pwa', 'web_pwa_oca')], merge_modules=False)
> env.cr.commit()

flachica:
> n [https://apps.odoo.com/apps/modules/12.0/web_pwa_oca/](url):

The link is broken?

FernandoRomera:
Or this
https://github.com/OCA/l10n-brazil/blob/12.0/l10n_br_base/migrations/12.0.1.0.1/pre-migration.py

flachica:
@FernandoRomera I include the script. Thanks

mgmtsystem_information_security/__manifest__.py
```diff
@@ -0,0 +1,31 @@
...
+    "version": "13.0.1.0.0",
+    "author": "Savoir-faire Linux, Odoo Community Association (OCA)",
+    "website": "http://www.savoirfairelinux.com",
+    "license": "AGPL-3",
+    "category": "Generic Modules/Others",
+    "depends": ["mgmtsystem_manual"],
+    "data": ["data/information_security_manual.xml"],
+    "demo": [],
+    "installable": True,
+}
```
FernandoRomera: Missing:
 "external_dependencies": {"python": ["openupgradelib"]},

And in requirements.txt:
openupgradelib

FernandoRomera (changes requested):

mgmtsystem_information_security/__manifest__.py
flachica: Thanks for revision. Included yer

mgmtsystem_information_security/__manifest__.py
```diff
@@ -0,0 +1,32 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    "name": "Information Security Management System Manual",
+    "version": "13.0.1.0.0",
+    "author": "Savoir-faire Linux, Odoo Community Association (OCA)",
+    "website": "http://www.savoirfairelinux.com",
```
ivantodorovich:
```suggestion
    "website": "https://github.com/OCA/management-system",
```

mgmtsystem_information_security/migrations/13.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,12 @@
...
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    openupgrade.update_module_names(
+        env.cr,
+        [("mgmtsystem_info_security_manual", "mgmtsystem_information_security")],
+        merge_modules=True,
+    )
```
ivantodorovich: This won't work, I think..

On current deployments `mgmtsystem_information_security` won't be installed, so this script will never run.

I'm not sure what the process is to handle this, most probably a script in https://github.com/OCA/openugprade ?

in any case, is it really worth the trouble to rename the module?

flachica: All right. I will rename ASAP. More simple is more good. Thanks @ivantodorovich

flachica: Renamed! Thanks

FernandoRomera (approved):

AntoniRomera (approved):

FernandoRomera:
@dreispt can you remove the request or review?

PR #327, opened 2020-04-22
[11.0] [FIX] fields related should be readonly True

nicomacr:
Only impact in V11.

If you are user with the low level of rights "Viewer" group
Try to create an nonconformity record, fill the fields an the try to save.
An error occurs related to stage model.

The problem is with this related fields that need to be read-only.

The desired behaviour is this user be able to create your own non conformity records.


pedrobaeza (approved):

pedrobaeza:
Fast-tracking this easy one

/ocabot merge patch

PR #328, opened 2020-04-23
[FIX] mgmtsystem: Changed label of model field name

luiseevaquer:
Changed to avoid warning in log due to have the same label with MRP module field name


pedrobaeza (approved):

hugho-ad (approved):

moylop260 (approved):
👍

PR #337, opened 2020-06-25
[FIX] Fix security, forms and template

Rad0van:
This fixes that upon installation of this all the other templates and documents get hidden (caused by security). Also changes display to the same forms as the rest. And also fixes content of the template being properly used.

@dreispt and @ivantodorovich could maybe have a look? Thanx.


ivantodorovich (approved):
I'm ok with it, to be honest I don't get why this security role was here.
But let's see what others say, maybe I'm missing something

Rad0van:
> I'm ok with it, to be honest I don't get why this security role was here.
> But let's see what others say, maybe I'm missing something

@flachica maybe?

flachica:
> > I'm ok with it, to be honest I don't get why this security role was here.
> > But let's see what others say, maybe I'm missing something
> 
> @flachica maybe?

This file was include [here](https://github.com/OCA/management-system/commit/42d04a675efd65eefd492deed14fe1bae8baf136)

Rad0van:
> > > I'm ok with it, to be honest I don't get why this security role was here.
> > > But let's see what others say, maybe I'm missing something
> > 
> > 
> > @flachica maybe?
> 
> This file was include [here](https://github.com/OCA/management-system/commit/42d04a675efd65eefd492deed14fe1bae8baf136)

@flachica but it seems the security for other document pages from that commit got lost somewhere in the process - for example document_page_procedure/security/procedure_security.xml, document_page_work_instruction/security/work_instruction_security.xml and mgmtsystem_manual/security/manual_security.xml. They no longer exist in 13.0 so I removed the last one as well.

flachica:
> > > > I'm ok with it, to be honest I don't get why this security role was here.
> > > > But let's see what others say, maybe I'm missing something
> > > 
> > > 
> > > @flachica maybe?
> > 
> > 
> > This file was include [here](https://github.com/OCA/management-system/commit/42d04a675efd65eefd492deed14fe1bae8baf136)
> 
> @flachica but it seems the security for other document pages from that commit got lost somewhere in the process - for example document_page_procedure/security/procedure_security.xml, document_page_work_instruction/security/work_instruction_security.xml and mgmtsystem_manual/security/manual_security.xml. They no longer exist in 13.0 so I removed the last one as well.

This is good for me

flachica (approved):

PR #339, opened 2020-06-25
[FIX] Fix automatic install of work instruction template

Rad0van:
This just adds document_page_work_instruction as dependency for document_page_procedure to get it automatically installed as in 10.0

### DIFF
--- a/mgmtsystem/models/res_config.py
+++ b/mgmtsystem/models/res_config.py
@@ -102,3 +102,8 @@ class MgmtsystemConfigSettings(models.TransientModel):
         help="Provide document approval and history. \n"
         "- This installs the module document_page_approval.",
     )
+    module_document_page_work_instruction = fields.Boolean(
+        "Work Instructions",
+        help="Provide Work Instructions category.\n"
+        "- This installs the module document_page_work_instruction.",
+    )
--- a/mgmtsystem/views/res_config.xml
+++ b/mgmtsystem/views/res_config.xml
@@ -243,6 +243,18 @@
                             </div>
                         </div>
 
+                        <div class="col-xs-12 col-md-6 o_setting_box">
+                            <div class="o_setting_left_pane">
+                                <field name="module_document_page_work_instruction" class="oe_inline"/>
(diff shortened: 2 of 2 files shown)


flachica (approved):

ivantodorovich:
If there isn't a real dependency, I'm not sure it makes sense to add it. The user can install both modules

Rad0van:
> If there isn't a real dependency, I'm not sure it makes sense to add it. The user can install both modules

@ivantodorovich all the document templates are installed through settings of the main module. This one does not have an option there. I was checking in older version how this was handled. In 10.0 it was dependency for document_page_procedure then it disappeared. So for the sake of completeness I opted to restore the dependency as in 10.0 Otherwise user would have to install it through main Apps management...

ivantodorovich:
Maybe we should add an option to the res.config page instead :thinking: , so that it's installed like the other modules

Rad0van:
> Maybe we should add an option to the res.config page instead 🤔 , so that it's installed like the other modules

That is an option as well. Should I do that?

ivantodorovich:
I think it does make more sense than adding the dependancy, and it's consistent to how the other modules are installed.
You can repurpose this PR for this matter

Rad0van:
> I think it does make more sense than adding the dependancy, and it's consistent to how the other modules are installed.
> You can repurpose this PR for this matter

@ivantodorovich OK, done. Should I squash the commits together? Also, #341 should be merged first I guess?

ivantodorovich:
Great! Thanks :)
Yes please squash them

Rad0van:
> Great! Thanks :)
> Yes please squash them

Hope I got it right...

ivantodorovich:
Not quite :sweat_smile: 
We need only 1 commit here

Rad0van:
I see only one. Am I wrong? Edit: I see now. Will fix.

Rad0van:
> Not quite 😅
> We need only 1 commit here

Done. I hope.

ivantodorovich (approved):

PR #341, opened 2020-06-25
[FIX] Fix the automatic un/install and status display

Rad0van:
Fixes un/installation of modules and proper display of their installation status. The attributes were misnamed so this was not working.


flachica (approved):

ivantodorovich (approved):

Rad0van:
@ivantodorovich if you could please also have a look at #337, #338, #339 and #340. Simple fixes there as well. Thanx a lot.

PR #344, opened 2020-07-08
[12.0][ADD] mgmtsystem_nonconformity_fieldservice

marcelsavegnago:
Add module mgmtsystem_nonconformity_fieldservice


etobella:
Just a question, this module depends on `fieldservice` that is somehow a vertical, shouldn't this module be there?

marcelsavegnago:
> Just a question, this module depends on `fieldservice` that is somehow a vertical, shouldn't this module be there?

@etobella I honestly believed it could be here. But without a problem, I can do it directly in the OCA/field-service project. Do you think better?. Management System is a vertical one too, isn't it?

etobella:
I know that could be in both sides, but I think that the good one is `fieldservice` because there is already accounting, stock and much more repos. IMO, `fieldservice` is a vertical and all the related modules needs to be directly there

Also, Management system is not a vertical, as it only manages an aspect that a lot of companies need to fulfill, from all sectors, `fieldservice` is for just some kind of companies, that's why I say it is a vertical (or at least, similar)

marcelsavegnago:
> I know that could be in both sides, but I think that the good one is `fieldservice` because there is already accounting, stock and much more repos. IMO, `fieldservice` is a vertical and all the related modules needs to be directly there

Ok I'll change and submit to the Field Service repository. Thanks for your help and explanation.

marcelsavegnago:
> I know that could be in both sides, but I think that the good one is `fieldservice` because there is already accounting, stock and much more repos. IMO, `fieldservice` is a vertical and all the related modules needs to be directly there
> 
> Also, Management system is not a vertical, as it only manages an aspect that a lot of companies need to fulfill, from all sectors, `fieldservice` is for just some kind of companies, that's why I say it is a vertical (or at least, similar)

@etobella https://github.com/OCA/field-service/pull/585

PR #282, opened 2020-01-25
[11.0] [ADD] new module mgmtsystem_action_template

mymage:
Ref. #263

PR #351, opened 2020-07-18
[12.0] [ADD] new module mgmtsystem_extend_action_template

mymage:
This is the migration of #282 still waiting review: I hope it can have more audience on this version :-)
Ref. #263


marcelsavegnago (approved):
Functional review Ok

etobella:
It works, but it is somehow crazy to use. 

Templates and actions are mixed and this is hard to use. Maybe you could inherit the base module and implement it in another class, this way it is easier to manage. WDYT?

jarroyomorales:
@mymage @etobella  In my opinion the best solution is to separate the action and the action.template model, I don't see any reason to mix them and makes us add an extra domain everywhere.
I could work on it if you are busy, we are also very interested in this. Since this module was a PR for v11 too and someone may be using it in production, we could try to do some migration scripts between the `is_template` version and the two models version.

jarroyomorales:
Hi @mymage ! I did a PR implementing the version with the `mgmtsystem.action.template` model here https://github.com/mymage/management-system/pull/1

What do you think?

mymage:
@etobella @jarroyomorales Thanks to give attention to this PR.
In recent time I don't worked on this "project" so I loose focus on it.
In the next days I will recap and reply to your notes.

mymage:
@etobella @jarroyomorales Hi, I did the changes on your suggestions.
Thanks @jarroyomorales for your PR: for me it was very useful on some aspects. Please give me reference to complete copyright/contributors (email, web).
As on other PR, I think that codecov is complaining for the lack of test (that I'm not able to do): I will tray later.

mymage:
@jarroyomorales Do you think that I have to update the 11.0 version PR too and then you will do the migration script?

mgmtsystem_action_template/tests/__init__.py
```diff
@@ -0,0 +1 @@
+from . import test_action_template
```
etobella: Newline

etobella (approved):

etobella:
@OCA/management-systems-maintainers I think this can be merged :smile: 
@mymage About the migration script, maybe, as the original PR has not been merged, you can apply the same changes there. I don't know if it has been used, so, if it hasn't, you shouldn't need a migration script.

mymage:
@etobella Thanks for review. I alligned the #282 with these changes.
A curiosity: is the codecov/project error resolvable? If yes can you give me some reference? Thanks

etobella:
You can add more coverage, but it is fine IMO, so you can leave it as it is 😄

PR #356, opened 2020-08-07
mgmtsystem_action: Action statusbar not clickable

flachica:
The Action form's statusbar is intended to be clickable, but Odoo 13 is apparently no longer compatible with clickable="True". Instead, either clickable="1" or options="{'clickable': 1}" must be used. Odoo core code uses the latter, so the Claim form has been updated to match core convention.


ivantodorovich (approved):

ivantodorovich:
Fastracking this

/ocabot merge patch

PR #357, opened 2020-08-07
mgmtsystem_nonconformity: Non conformity statusbar not clickable

flachica:
The Non conformity form's statusbar is intended to be clickable, but Odoo 13 is apparently no longer compatible with clickable="True". Instead, either clickable="1" or options="{'clickable': 1}" must be used. Odoo core code uses the latter, so the Claim form has been updated to match core convention.


ivantodorovich:
Can you squash the commits?

flachica:
@ivantodorovich Squashed. I forgot increase version in manifest

ivantodorovich (approved):

ivantodorovich:
Fast-tracking this

/ocabot merge nobump

PR #372, opened 2021-02-22
[MIG] mgmtsystem_nonconformity: Migration to 14.0

dzungtran89:
Dependencies:
- mgmtsystem_action: https://github.com/OCA/management-system/pull/366
- document_page_procedure: https://github.com/OCA/management-system/pull/369

PR #367, opened 2021-02-13
[MIG] mgmtsystem_survey: Migration to 14.0

dzungtran89:
Dependencies:
- mgmtsystem: https://github.com/OCA/management-system/pull/365

PR #373, opened 2021-02-22
[MIG] mgmtsystem_review: Migration to 14.0

dzungtran89:
Dependencies:
- [x] mgmtsystem_nonconformity: https://github.com/OCA/management-system/pull/372
- [x] mgmtsystem_survey: https://github.com/OCA/management-system/pull/367


YvanDotet:
Hello
Is it ready to merge in the 14.0 branch ?
Thank you.

test-requirements.txt
```diff
@@ -0,0 +1,7 @@
+odoo14-addon-knowledge @git+https://github.com/dzungtran89/knowledge@14.0-mig-knowledge#subdirectory=setup/knowledge
```
simahawk: to cleanup

simahawk (changes requested):
LG

gurneyalex:
as the dependencies are now merged, can you rebase to trigger a rebuild?

dzungtran89:
Hi @gurneyalex , the PR has been updated

mgmtsystem_review/models/mgmtsystem_review.py
```diff
@@ -6,6 +6,7 @@
 
 class MgmtsystemReview(models.Model):
     _name = "mgmtsystem.review"
+    _inherit = ["mail.thread", "mail.activity.mixin"]
```
simahawk, line 9: for this change you should add the dependency on `mail` no?

hailangvn: Good day @simahawk, thanks for your comment. This module depends on `mgmtsystem_survey`, which depends on `survey`, and in turn, depends on `mail`. So we can keep this as is.

sebalix (approved):
LG

mgmtsystem_review/models/mgmtsystem_review.py
simahawk: I always state such direct dependencies explicitly but that's just my pov ;)

simahawk (approved):

PR #429, opened 2022-05-23
[13.0][MIG] mgmtsystem action efficacy

mymage:
Migration to 13.0


mgmtsystem_action_efficacy/tests/test_efficacy.py
```diff
@@ -0,0 +1,23 @@
...
+class TestActionEfficacy(common.TransactionCase):
+    def test_change_efficacy(self):
+
+        record = self.env["mgmtsystem.action"].search([])[0]
+
+        record.efficacy_value = 50
+        record._onchange_efficacy_value()
+        self.assertEqual(50, record.efficacy_value)
+
+        with self.assertRaises(ValidationError) as e:
```
ivantodorovich, line 19: Non blocking tips: 

You can use `self.assertRaisesRegex` like this:

```py
with self.assertRaisesRegex(ValidationError, "Rating must be between 0 and 100"):
    ...
```


Also to test onchanges it's better to use `odoo.tests.common.Form`. This way you don't have to explicitly call the `_onchange_*` method.

ivantodorovich (approved):

mgmtsystem_action_efficacy/tests/test_efficacy.py
mymage: > Non blocking tips:
> 

Thanks for suggestions. I'm not confident with tests so every information is useful. I will try with the 14.0 migration.

etobella (approved):

PR #436, opened 2022-08-11
[15.0] [IMP] `mgmtsystem`: bump app icon to 15.0 style

ivantodorovich:
Specially for the `mgmgsystem` module (which has an app menu)


pedrobaeza:
Thanks for the initiative. But please don't put the exact same icon for all the modules. They should have any little difference on each one, like overlapping other little icon for each submodule.

ivantodorovich:
> Thanks for the initiative. But please don't put the exact same icon for all the modules. They should have any little difference on each one, like overlapping other little icon for each submodule.

Please note they all had the exact same icon before this PR (with the exception of `mgmtsystem_nonconformity_hr`, but it's not relevant IMO)

Also that there's only 1 icon that matters: the one for `mgmtsystem`, because it's the one used as application icon in the menu item. There's close to 0 value on designing or adding icons to the other modules. I don't think it's worth spending time on them.

Having said that maybe it's a good idea to simply remove the icons (and keep the default OCA one) for all modules except `mgmtsystem`. That would avoid the duplicated icon binaries in the repo (and the maintenance work to keep them all in sync). What do you think?

pedrobaeza:
Well, such icon is shown for example in the OCA apps store, so I prefer to have them with the same general icon but a different subicon, as I did for example in vertical-association (oldish style, as this was done on 8.0):

https://github.com/OCA/vertical-association/blob/14.0/membership_extension/static/description/icon.svg
https://github.com/OCA/vertical-association/blob/14.0/membership_prorate/static/description/icon.svg
https://github.com/OCA/vertical-association/blob/14.0/membership_prorate_variable_period/static/description/icon.svg

I can help you on this if you want.

ivantodorovich:
I've changed the scope of the PR to limit it to only the `mgmtsystem` module, that's the one I really care about anyway (for the menu icon)

I disagree about having to design app icons for each non-application module. Designing icons is tedious and adds little to no value if it's not displayed in a menu. Also.. Imagine having to do that for all OCA modules if that were the convention 😩 there probably aren't enough meaningful glyphs out there for all of them. We'd start repeating very quickly.

You mentioned the OCA apps store (and I guess we should include the Odoo app store, too), but I think that should be dealt with differently. It's not really the app icon you want to show there, but rather the marketing banner. And the marketing banner is not limited to a glyph, it can even be a gif.. so it offers more information for the people trying to discover a new module.

See for example the odoo app store:
image

And compare with other modules' banners you see on their home page (most paid top-selling modules have one):
https://apps.odoo.com/apps/modules/browse

pedrobaeza (approved):
OK, I may do the rest of the derivatives when merged

ivantodorovich:
> OK, I may do the rest of the derivatives when merged

Okay then :)
I made this tool to ease the process if your interested: https://ivantodorovich.github.io/odoo-icon/


/ocabot merge patch

PR #462, opened 2022-12-13
[16.0][MIG] document_page_environment_manual: Migration to 16.0

FernandoRomera:


.pre-commit-config.yaml
```diff
@@ -27,6 +27,11 @@ repos:
         entry: found forbidden files; remove them
         language: fail
         files: "\\.rej$"
+      - id: en-po-files
+        name: en.po files cannot exist
+        entry: found a en.po file
+        language: fail
+        files: '[a-zA-Z0-9_]*/i18n/en\.po$'
```
ivantodorovich, line 34: These changes shouldn't be in the `[MIG] document_page_environment_manual: Migration to 16.0` commit.
It should be a separate `[UPD] update copier dotfiles`

ivantodorovich (changes requested):

PR #470, opened 2022-12-14
[16.0][MIG] mgmtsystem_audit: Migration to 16.0

FernandoRomera:


AntoniRomera (approved):

MRomeera (approved):
Perfect!

FernandoRomera:
@max3903 
Once the previous PR (mgmtsystem_nonconformity) has been merged, we can proceed with this one.

mgmtsystem_audit/data/audit_automated_actions.xml
```diff
@@ -33,7 +34,7 @@ Odoo
         <field name="model_id" ref="model_mgmtsystem_audit" />
         <field name="template_id" ref="template_email_prepare_to_audit" />
         <field name="trigger">on_time</field>
-        <field name="state">email</field>
+        <!-- <field name="state">email</field> -->
```
ivantodorovich:
```xml
<field name="state">mail_post</field>
```

I think you want `mail_post` here, no?
Commenting it out will fallback to the default `object_write`, and that's not what we want here

See: https://github.com/odoo/odoo/commit/b3fc41cc44b251be37b6aaf2931179afeb126d69

ivantodorovich (changes requested):

mgmtsystem_audit/data/audit_automated_actions.xml
ivantodorovich: (same for the other case, btw)

FernandoRomera: @ivantodorovich 
Thanks.

ivantodorovich (approved):

PR #471, opened 2022-12-14
[16.0][MIG] mgmtsystem_nonconformity: Migration to 16.0

FernandoRomera:


mgmtsystem_nonconformity/views/mgmtsystem_action.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
max3903, line 1:
```suggestion
```

There is no need to add those lines. Thanks!

FernandoRomera: The documentation says: 
https://www.odoo.com/documentation/16.0/es/developer/reference/backend/data.html
for all XML files.

pedrobaeza: On contrary, Maxime, they must be there. XML format specifies that. If they are not there, they are working by chance, and there's the possibility of a future version of `lxml` (or the low-level library used by Odoo in the future) doesn't parse it correctly. There can be also problems with certain characters if the encoding is not specified.

ivantodorovich (approved):

AntoniRomera (approved):

FernandoRomera:
@max3903 
Can you merge this PR, please?

mgmtsystem_nonconformity/views/mgmtsystem_action.xml
max3903: Ok

PR #479, opened 2022-12-17
[15.0][ADD] test mgmtsystem_nonconformity_quality_control_oca

mymage:


mymage:
@ivantodorovich as promised, the tests

max3903 (approved):

mgmtsystem_nonconformity_quality_control_oca/tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
```
ivantodorovich:
```suggestion
```

nitpicking: there's no need to put headers on __init__.py files

mgmtsystem_nonconformity_quality_control_oca/tests/test_inspection.py
```diff
@@ -0,0 +1,61 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import logging
+
+from odoo.tests import common
+
+logger = logging.getLogger(__name__)
```
ivantodorovich:
```suggestion
from odoo.tests import common
```

Seems `logger` is not used

mgmtsystem_nonconformity_quality_control_oca/tests/test_inspection.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+class TestQualityControl(common.TransactionCase):
+    def setUp(self):
```
ivantodorovich:
```suggestion
    def setUpClass(cls):
```
(and replace all occurrences of `self` with `cls` within this method)



Keep in mind `setUp` runs before every test.
whilst `setUpClass` runs only once for all tests, so it's more performant

setup/_metapackage/VERSION.txt
```diff
@@ -1 +1 @@
-15.0.20221214.0
```
ivantodorovich: What's this change? 🤔

setup/_metapackage/setup.py
```diff
@@ -26,6 +26,7 @@
         'odoo-addon-mgmtsystem_nonconformity_hr>=15.0dev,<15.1dev',
         'odoo-addon-mgmtsystem_nonconformity_mrp>=15.0dev,<15.1dev',
         'odoo-addon-mgmtsystem_nonconformity_product>=15.0dev,<15.1dev',
+        'odoo-addon-mgmtsystem_nonconformity_quality_control_oca>=15.0dev,<15.1dev',
```
ivantodorovich: Same here. Weird. We shouldn't update this manually I think

mgmtsystem_nonconformity_quality_control_oca/tests/__init__.py
mymage: ok, copy that

mgmtsystem_nonconformity_quality_control_oca/tests/test_inspection.py
mymage: Yes, I forget to delete it

mgmtsystem_nonconformity_quality_control_oca/tests/test_inspection.py
mymage: I tried but pre-commit dislike it. So I changed setUp but not cls.
Do you have some other suggestion?
Yes: add @classmethod before the method definition ;-)

setup/_metapackage/VERSION.txt
mymage: See bottom

setup/_metapackage/setup.py
mymage: Probably I did something wrong. I did a rebase if this can fix things.

ivantodorovich (approved):

PR #495, opened 2022-12-26
[16.0][ADD] oca_dependencies.txt

mymage:
I think that this file had to be add.
Looking into mgmtsystem_claim I suppose that crm should be included.
I don't know if there are other repositories to add.


ivantodorovich:
Originally it was needed for our CI, but  it's not needed anymore, as dependencies are installed with `pip`.

PR #496, opened 2022-12-26
[16.0][MIG] mgmtsystem_claim

mymage:
depends https://github.com/OCA/crm/pull/468


mgmtsystem_claim/views/res_partner_views.xml
```diff
@@ -17,7 +13,7 @@
                     context="{'search_default_partner_id': active_id, 'default_partner_id': active_id}"
                     name="%(mgmtsystem_claim_action)d"
                     icon="fa-comments"
-                    groups="sales_team.group_sale_salesman"
+                    groups="sales_team.group_sale_salesman,base.group_partner_manager"
```
ivantodorovich: Don't add the `base.group_partner_manager` here. It'll change the previous behaviour.
I don't know why it was added to the view `groups_id` before, but only users belonging to `sales_team.group_sale_salesman` were able to see this button (as it was the only group defined in the element's `groups` attribute). So it shuold remain like this, IMO

mymage: Following  the migration instruction, I deleted
```
<field
            name="groups_id"
            eval="[(4, ref('sales_team.group_sale_salesman')), (4, ref('base.group_partner_manager'))]"
        />
```
adding the "base.group_partner_manager" to the already present groups parameter.
I'm not confident with this change, so if you are sure, I trust you and delete it.

ivantodorovich: Yes, please.
The explaination is this:


**Before:**

- The `ir.ui.view` was applied for both groups
- The button was shown only for `sales_team.group_sale_salesman`
- _Effectively, only users belogning to `sales_team.group_sale_salesman` saw the button_

**With your change:**

- The `ir.ui.view` is applied for all groups (normal, as Odoo deprecated the use of `groups_id`)
- The button is shown for both `sales_team.group_sale_salesman` and `base.group_partner_manager`


---

As you can see, the behaviour has changed.
To keep the previous behaviour we simply need to remove the `group_id` from the `ir.ui.view`, and keep the previous value for `groups` on the button element

mymage: Thanks for explanation. ASAP I will fix tests using "remote" pr on crm.

primes2h (changes requested):
This change must be moved in the migration commit.

mgmtsystem_claim/views/res_partner_views.xml
```diff
@@ -13,7 +13,7 @@
                     context="{'search_default_partner_id': active_id, 'default_partner_id': active_id}"
                     name="%(mgmtsystem_claim_action)d"
                     icon="fa-comments"
-                    groups="sales_team.group_sale_salesman,base.group_partner_manager"
+                    groups="sales_team.group_sale_salesman"
```
primes2h: This change must be moved in the migration commit

test-requirements.txt
```diff
@@ -0,0 +1,2 @@
+# odoo-addon-<module_name_with_dashes> @ git+https://github.com/OCA/<repository>/refs/pull/<PR_number>/head#subdirectory=setup/<module_name>
+odoo-addon-crm_claim @ git+https://github.com/OCA/crm/refs/pull/468/head#subdirectory=setup/crm_claim
```
primes2h:
```suggestion
odoo-addon-crm_claim @ git+https://github.com/OCA/crm/pull/468/head#subdirectory=setup/crm_claim
```

primes2h (changes requested):

mgmtsystem_claim/views/res_partner_views.xml
mymage: Yes I know, but I'm lazy ;-)

ivantodorovich: Still, he makes a good point and it should be fixed 😅 
It's currently in the `[DONT MERGE] test-requirements.txt` commit, which will be removed after the dependency is merged. We don't want to loose these changes by mistake so better to put them in the right commit 😄

mymage: ok guys, done

test-requirements.txt
mymage: the solution was:
odoo-addon-crm_claim @ git+https://github.com/OCA/crm.git@refs/pull/468/head#subdirectory=setup/crm_claim

FernandoRomera:
@mymage 
Done https://github.com/OCA/crm/pull/468

dreispt (approved):

etobella:
You removed groups_id on a view. Might be necessary to add a migration script, as it will not be changed when migrating on old databases, making two different behaviours

dreispt:
@mymage Maybe you need to add the  `<t groups="...">` element?

mymage:
@etobella you have too much faith in my abilities :-) but if it the case I'll learn how to do it
@dreispt I hope I have interpreted your suggestion in the right way and that this avoids the route suggested by @etobella

ivantodorovich:
> You removed groups_id on a view. Might be necessary to add a migration script, as it will not be changed when migrating on old databases, making two different behaviours

AFAICS there's no migration script needed.

Before, the view had the field `groups_id` set, it's true.
This is a field Odoo deprecated. We're no longer able to set groups on ir.ui.view records
We must set the groups on the view's arch elements, instead.

However, the button element already had a group defined: `groups="sales_team.group_sale_salesman"`
So in fact the groups in the view record was just redundant

ivantodorovich (approved):

primes2h:
@mymage 
Please merge/squash commits generated by bots. See https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate

PR #517, opened 2023-03-24
[12.0][MIG] mgmtsystem_nonconformity_quality_control_oca: Migration to 12.0

kaynnan:


max3903:
@kaynnan `quality_control` is an enterprise module. You cannot push this module to the OCA. You can still publish it in your own public repo... or remove the dependency.

marcelsavegnago:
> @kaynnan `quality_control` is an enterprise module. You cannot push this module to the OCA. You can still publish it in your own public repo... or remove the dependency.

@max3903 in 14.0 the module mgmtsystem_nonconformity_quality_control_oca depends on quality_control_oca. Already in version 12.0 it is called quality_control https://github.com/OCA/manufacture/tree/12.0/quality_control

I understand that the module name was changed to _oca because of this name conflict with the Enterprise module. Better not to merge this PR in the management-system ?

max3903:
@marcelsavegnago As I understand it, we already have a name conflict with https://github.com/OCA/manufacture/tree/12.0/quality_control. 

I would rename/duplicate the module to fix it first so the dependencies are manageable and mgmtsystem_nonconformity_quality_control_oca can then be hosted here.

marcelsavegnago:
> I would rename/duplicate the module to fix it first so the dependencies are manageable and mgmtsystem_nonconformity_quality_control_oca can then be hosted here.

I thought about renaming. I will organize this task here and soon we will send a PR

PR #519, opened 2023-04-13
[14.0] Use tags to detect procedures

### LINKED ISSUE #518: Improvement on detection of Procedures
Right now, the only way to detect procedures is to use a single category created as data. It is a problem when the environment is complex, as a simple category for all your procedures might be hard to manage.

I have two ideas for that:

1- Adding them as `type`of document.page (some changes might be necessary, but not too much)
2- Create a subtype

Obviously, some migrations scripts are necessary, but I might be able to do that :smile: 

This could be used for procedures, work instructions, quality manuals...

@max3903 WDYT?

etobella:
Includes migration scripts

### DIFF
--- a/document_page_environmental_aspect/__manifest__.py
+++ b/document_page_environmental_aspect/__manifest__.py
@@ -25,7 +25,11 @@
     "website": "https://github.com/OCA/management-system",
     "license": "AGPL-3",
     "category": "Generic Modules/Others",
-    "depends": ["document_page", "mgmtsystem"],
-    "data": ["data/document_page.xml", "views/document_page.xml"],
+    "depends": ["document_page_tag", "mgmtsystem"],
+    "data": [
+        "data/document_page.xml",
+        "data/document_page_tag.xml",
+        "views/document_page.xml",
+    ],
     "development_status": "Production/Stable",
 }
--- a/document_page_environmental_aspect/data/document_page.xml
+++ b/document_page_environmental_aspect/data/document_page.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" ?>
-<odoo>
+<odoo noupdate="1">
+
     <record id="document_page_group_environmental_aspect" model="document.page">
       <field name="name">Environmental Aspect</field>
       <field name="type">category</field>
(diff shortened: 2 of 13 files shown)


etobella:
@max3903

max3903:
@etobella I meant to use tags but also to keep the category since it also provides the content template.

document_page_environmental_aspect/data/document_page.xml
```diff
@@ -1,26 +1,8 @@
...
-
-<h1>Responsibilities</h1>
-
-<h1>Objectives</h1>
-
-<h1>Programs</h1>
-        ]]>
-      </field>
+    <record id="environment_aspect_tag" model="document.page.tag" forcecreate="1">
+        <field name="name">Environment Aspect</field>
```
max3903: @etobella I would keep the original file as it was and add a new one `document_page_tag.xml` with this content.

document_page_environmental_aspect/__manifest__.py
```diff
@@ -25,7 +25,11 @@
     "website": "https://github.com/OCA/management-system",
     "license": "AGPL-3",
     "category": "Generic Modules/Others",
-    "depends": ["document_page", "mgmtsystem"],
-    "data": ["data/document_page.xml", "views/document_page.xml"],
+    "depends": ["document_page_tag", "mgmtsystem"],
+    "data": [
+        "data/document_page.xml",
+        "data/document_tag.xml",
```
max3903:
```suggestion
        "data/document_page_tag.xml",
```

document_page_procedure/__manifest__.py
```diff
@@ -3,13 +3,17 @@
...
     "author": "Savoir-faire Linux, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/management-system",
     "license": "AGPL-3",
     "category": "Management System",
-    "depends": ["document_page", "mgmtsystem"],
-    "data": ["data/document_page_procedure.xml", "views/document_page_procedure.xml"],
+    "depends": ["document_page_tag", "mgmtsystem"],
+    "data": [
+        "data/document_page_procedure.xml",
+        "data/document_tag.xml",
```
max3903:
```suggestion
        "data/document_page_tag.xml",
```

max3903 (approved):

sanchonuria (approved):

PR #524, opened 2023-06-13
[FIX] mgmtsystem_review: Wrong mig script data version

fuentes73:
The migration script is pointing to a wrong path


etobella:
Can you reabase?

etobella:
/ocabot merge patch
I will try to merge it directly, as the change should be correct

etobella (approved):

PR #528, opened 2023-07-25
[FIX] mgmtsystem_review: post-migration.py

angelmoya:


etobella (approved):
/ocabot merge patch
Fast tracking, as migration would fail

etobella:
@angelmoya I aborted it for two reasons:
1- Commit message is not clean enough
2- There is a previous PR fixing this too

max3903 (approved):

PR #531, opened 2023-09-13
Update mgmtsystem_nonconformity_cause.py

yassine-rhimi:
add unaccent=False


yassine-rhimi:
why this MR is not merging yet? Is there a problem?

PR #533, opened 2023-09-14
[IMP] mgmtsystem_audit: mgmtsystem_nonconformity.py

yassine-rhimi:
The declaration of the function create in v16 will be decorated like this:

@api.model_create_multi
def create(self, vals):


yassine-rhimi:
why this MR is not merging yet? Is there a problem?

yassine-rhimi:
why this MR is not merging yet? Is there a problem?

max3903:
@yassine-rhimi Can you split your changes to have one pr/branch per module?

PR #541, opened 2023-10-24
[FW][13.0] nonconformity_stage : state field changed to non ReadOnly

grindtildeath:
FW Port of #435 

Depends on:
 - [ ] #545


simahawk:
you can rebase

grindtildeath:
@simahawk done

PR #558, opened 2023-12-05
[14.0][FIX] mgmtsystem_evaluation: fix domain

kluna1998:
i think that to reference the value of a current record is without quotes


etobella (approved):
Fast tracking as this is a bug,

Thanks for finding it

/ocabot merge patch

PR #566, opened 2024-01-18
[16.0] [IMP] mgmtsystem_nonconformity: Show non conformities on all records

etobella:
Before this change, we needed to add patches everywhere. With this change we just need to click on the non-conformity button on the record chatter. It does not appear if you don't have access to mgmtsystem.

Captura desde 2024-01-18 21-47-33

### DIFF
--- a/mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
+++ b/mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
@@ -43,7 +43,9 @@ class MgmtsystemNonconformity(models.Model):
     closing_date = fields.Datetime(readonly=True)
 
     partner_id = fields.Many2one("res.partner", "Partner", required=True)
-    reference = fields.Char("Related to")
+    reference = fields.Char(
+        "Related to", default=lambda self: self._default_reference()
+    )
     responsible_user_id = fields.Many2one(
         "res.users", "Responsible", required=True, tracking=True
     )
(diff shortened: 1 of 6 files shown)


mcassuto:
Brilliant ! Thanks @etobella

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity_abstract.py
```diff
@@ -38,3 +40,50 @@ def action_view_non_conformities(self):
...
+            all_models = res["models"].copy()
+            for node in doc.xpath("/form/div[hasclass('oe_chatter')]"):
+                # _add_tier_validation_label process
+                new_node = etree.fromstring(
+                    "<field name='non_conformity_count' invisible='1'/>"
+                )
+                new_arch, new_models = View.postprocess_and_fields(new_node, self._name)
+                new_node = etree.fromstring(new_arch)
+                for model in new_models:
+                    if model in all_models:
```
ValentinVinagre: Wouldn't it be better to perform a filter/ed in for and so you don't pivot on this condition ´if model in all_models:´?

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
```diff
@@ -137,9 +139,20 @@ def _stage_groups(self, stages, domain, order):
...
     )
-    res_model = fields.Char()
+    res_model = fields.Char(index=True)
     res_id = fields.Integer(index=True)
 
+    @api.model
+    def _default_reference(self):
+        if self.env.context.get("active_model") and self.env.context.get("active_id"):
+            return (
+                self.env[self.env.context["active_model"]]
```
ValentinVinagre, line 149: better with get?

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity_abstract.py
```diff
@@ -38,3 +40,50 @@ def action_view_non_conformities(self):
...
+
+    @api.model
+    def _get_view_fields(self, view_type, models):
+        """
+        We need to add this in order to fix the usage of form opening from
+        trees inside a form
+        """
+        result = super()._get_view_fields(view_type, models)
+        if view_type == "form":
+            if self.env.user.has_group("mgmtsystem.group_mgmtsystem_viewer"):
```
ValentinVinagre: better join it with the if before ?

ValentinVinagre (approved):
Some review. Are you planning to add tests for this new imp?

mgmtsystem_nonconformity/models/mgmtsystem_nonconformity.py
etobella: No need, as we already checked if it existed on the if :wink:

PR #570, opened 2024-02-07
[IMP] mgmtsystem_nonconformity: Update comment fields' colspan

oa-eska:


dreispt:
@oa-eska Can you explain why the change? AFAIK colspan=2 is already the default.

PR #573, opened 2024-02-23
[17.0] [MIG] document_page_work_instruction

ThinkopenDev:
Straightforward migration from 16.0


mymage:
@ThinkopenDev please re-push so runboat will go green.

ThinkopenDev:
Sorry, closed PR on accident.

mymage (approved):
LGTM. If you can, allign the title with the standard, thanks

mymage:
While testing on Runboat I realize that the behavior is not the one described in the readme: the Categories item does not appear in the Configuration menu.
The entry is probably added from another module.
In this case for me it is better to change the readme. I don't know if it will be correct to add that module as a dependency.
If anyone has the time and knowledge to check, they are welcome.

mymage:
Hi @ThinkopenDev, the repo was fixed (see #585).
Then I think you have to re-push this after a rebase. Thanks!

max3903 (approved):

PR #589, opened 2024-05-21
[17.0] [MIG] mgmtsystem_action

m1k3lm:


m1k3lm:
@mymage should the test coverage be fixed in the same PR as the migration?

mymage:
> @mymage should the test coverage be fixed in the same PR as the migration?

I don't know really what is the actual policy. 
I had similar conditions in other migration PRs and I did my best to have all green.
If I remembe well some PSC merged it with some CodeCov not green.
So if you can and it is "simple" to complete the CodeCov, do it. If you do not reach the result, place here a message to [@]OCA/management-systems-maintainers.

mymage:
@m1k3lm ... but the pandoc file is still here :-(

m1k3lm:
> @m1k3lm ... but the pandoc file is still here :-(

Sorry, @mymage already deleted.

mymage (approved):
LGTM

max3903 (approved):

PR #631, opened 2025-01-20
[18.0][MIG] mgmtsystem_nonconformity_quality_control_oca: Migration to 18.0

BhaveshHeliconia:
depends on

- https://github.com/OCA/manufacture/pull/1374


mgmtsystem_nonconformity_quality_control_oca/views/qc_inspection.xml
```diff
@@ -23,11 +23,11 @@
                         />
                         <span
                             class="o_stat_text"
-                            attrs="{'invisible': [('mgmtsystem_nonconformity_count', '&lt;', 2)]}"
+                            invisible="[('mgmtsystem_nonconformity_count', '&lt;', 2)]"
```
ivantodorovich: This needs to be a pythonic expression now, not a domain

mgmtsystem_nonconformity_quality_control_oca/views/qc_inspection.xml
```diff
@@ -23,11 +23,11 @@
...
                         > Nonconformities</span>
                         <span
                             class="o_stat_text"
-                            attrs="{'invisible': [('mgmtsystem_nonconformity_count', '&gt;', 1)]}"
+                            invisible="[('mgmtsystem_nonconformity_count', '&gt;', 1)]"
```
ivantodorovich: same

ivantodorovich (changes requested):
/ocabot migration mgmtsystem_nonconformity_quality_control_oca

Hey, thanks for your contribution

Could you please check my comments below?
Odoo now evaluates pythonic expressions for `invisible` / `readonly` attributes, instead of domains.

With the current approach, the domain is a list that evaluates to `True`; thus it'd be the same as writing `invisible="True"`. It needs to be corrected

mgmtsystem_nonconformity_quality_control_oca/views/qc_inspection.xml
BhaveshHeliconia: @ivantodorovich, Thanks for the suggestion! it's fixed.

PR #670, opened 2025-03-27
[FIX] mgmtsystem_evaluation: support model_field_char option in reference widget to avoid access to ir.model

kluna1998:
fixing:

- AttributeError: 'mgmtsystem.evaluation.template' object has no attribute '_get_mail_thread_data'

- ValueError: Invalid field mail.activity.type.res_model_id in leaf ('res_model_id', '=', 421)

- reference field configured with model_field fails to render due to access denied errors when trying to read ir.model -> _fetchModelFieldReference using BasicModel.include, and add support for model_field_char as a fallback.

@etobella

### DIFF
--- a/mgmtsystem_evaluation/__manifest__.py
+++ b/mgmtsystem_evaluation/__manifest__.py
@@ -20,4 +20,9 @@
         "data/cron.xml",
     ],
     "demo": ["demo/demo.xml"],
+    "assets": {
+        "web.assets_backend": [
+            "mgmtsystem_evaluation/**/*",
+        ],
+    },
 }
--- a/mgmtsystem_evaluation/models/mgmtsystem_evaluation.py
+++ b/mgmtsystem_evaluation/models/mgmtsystem_evaluation.py
@@ -26,7 +26,10 @@ class MgmtsystemEvaluation(models.Model):
         string="Model technical name",
     )
     model_id = fields.Many2one(
-        "ir.model", compute="_compute_template_fields", store=True, readonly=False
+        "ir.model",
+        compute="_compute_template_fields",
+        store=True,
+        readonly=False,
     )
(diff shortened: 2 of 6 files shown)


etobella (commented):
The changes are correct, but if I access with demo user and minimal permissions I get the following error:

image

I don't know if  it is related :cry:

mgmtsystem_evaluation/security/ir.model.access.csv
```diff
@@ -5,3 +5,4 @@
 "manage_mgmtsystem_evaluation_template","mgmtsystem.evaluation.template","model_mgmtsystem_evaluation_template","mgmtsystem.group_mgmtsystem_user",1,1,1,0
 "access_mgmtsystem_evaluation_result","mgmtsystem.evaluation.result","model_mgmtsystem_evaluation_result","base.group_user",1,0,0,0
 "manage_mgmtsystem_evaluation_result","mgmtsystem.evaluation.result","model_mgmtsystem_evaluation_result","mgmtsystem.group_mgmtsystem_user",1,1,1,0
+"manage_mgmtsystem_evaluation_ir_model","ir.model","model_ir_model","mgmtsystem.group_mgmtsystem_user",1,0,0,0
```
kluna1998: I wasn’t able to fix the issue without giving read access to `ir.model` for the `group_mgmtsystem_user`.  
The problematic fields appear to be `model_id` and `resource`.  
Do you happen to know a cleaner solution that avoids granting this permission? 
@etobella

etobella: This is not an option

etobella (approved):

etobella:
Well, this is raising errors, and it is a problem on production environments

/ocabot merge patch

PR #679, opened 2025-05-26
[16.0] [IMP] mgmtsystem_hazard: Views and menu entry for mgmtsystem.hazard.hazard

luisDIXMIT:
Before this, we couldn't create or edit mgmtsystem.hazard.hazard manually.

@etobella

### DIFF
--- a/mgmtsystem_hazard/README.rst
+++ b/mgmtsystem_hazard/README.rst
@@ -54,6 +54,7 @@ Known issues / Roadmap
 ======================
 
 * Add multi-system and multi-company support
+* Remove the mgmtsystem.hazard.hazard model and any dependencies that other modules may have with it.
 
 Bug Tracker
 ===========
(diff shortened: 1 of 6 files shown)


luisDIXMIT:
I've been playing around with this model, and IMHO it isn't very useful. Maybe we could remove it. WDYT @max3903 ?

etobella:
@max3903 before removing it, we prefered to show the model, but I don't understand why we have a hazard.hazard inside a hazard (it looks something like a kind or type, but we alread have that :sob: )

max3903:
@etobella This module was made years ago for a demo. I didn't gethe project. I used the same model architecture of a different software given by the customer. I don't have any strong feeling about making changes and removing it.

etobella:
I will wait a few days, but I would opt for deletion of the field, as it has no sense.

luisDIXMIT:
After an internal discussion, we suggest for now to add the views in order to improve the model and make it more useful, as I explained in my initial comment. We also detected dependencies between other modules, so it is not advisable to delete the mgmtsystem.hazard.hazard model at this time. However, we agreed that the model's presence no longer makes much sense, so it is recommended to delete it in the future, as we discussed @max3903 . Therefore, we will add this to the roadmap.

etobella (approved):

etobella:
it is better to leave it in stable versions and remove the logic on 19. It takes time, but it is safer.

dreispt (approved):

PR #692, opened 2025-07-23
[15.0][IMP] mgmtsystem_review: Separate the survey part in a new module (mgmtsystem_review_survey)

victoralmau:
Separate the survey part in a new module (`mgmtsystem_review_survey`)

Before, if you needed to install `mgmtsystem_review` (because it was a dependency of `mgmtsystem_quality`) and you didn't use or want to use survey it was "mandatory", now, the new module `mgmtsystem_review_survey` adds the survey part to make that part optional.

@Tecnativa TT57340


mgmtsystem_review/report/review.xml
```diff
@@ -77,40 +77,6 @@
                 </div>
                 <div class="row" style="padding-top:10px">
                     <div class="col-12">
```
pedrobaeza, line 79: Put a name to the div for better xpath.

mgmtsystem_review_survey/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,3 @@
+* `Tecnativa <https://www.tecnativa.com>`_:
+
+  * Víctor Martínez
```
pedrobaeza: You have to put here the original contributors.

pedrobaeza (approved):
This split is very needed for correct modularity.

mgmtsystem_review/report/review.xml
victoralmau: Adding a name to the div would not work because what I want is to add **before** `<t t-foreach=“doc.line_ids” t-as=“line”>` what is related to `response_ids`.

pedrobaeza: It's a good part to always name anonymous elements like this, `th`, `td`, etc, but as you prefer.

mgmtsystem_review/README.rst
```diff
@@ -17,17 +13,17 @@ Management System - Review
 .. |badge1| image:: https://img.shields.io/badge/maturity-Beta-yellow.png
     :target: https://odoo-community.org/page/development-status
     :alt: Beta
-.. |badge2| image:: https://img.shields.io/badge/license-AGPL--3-blue.png
+.. |badge2| image:: https://img.shields.io/badge/licence-AGPL--3-blue.png
     :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
     :alt: License: AGPL-3
 .. |badge3| image:: https://img.shields.io/badge/github-OCA%2Fmanagement--system-lightgray.png?logo=github
-    :target: https://github.com/OCA/management-system/tree/15.0/mgmtsystem_review
+    :target: https://github.com/OCA/management-system/tree/16.0/mgmtsystem_review
```
pedrobaeza: 16.0?

victoralmau: That's strange, pre-commit has changed it. I see that https://github.com/OCA/management-system/blob/507d371f8aa657498e26b2faf860a961034a8f31/.pre-commit-config.yaml#L51 is at 16.0, it seems to come from https://github.com/OCA/management-system/commit/93dde48fe5a722e2b61ab3b538d427068c46283d

Should I change this in a separate PR to leave it “already fixed”?

pedrobaeza: It seems @dreispt mixed branches. I'm fixing it in 0a0f25309a1055c159a7d76cb5890e4760046fca. Please rebase and run `pre-commit run -a`.

mgmtsystem_review_survey/report/review.xml
```diff
@@ -0,0 +1,36 @@
...
+    <template id="review_report" inherit_id="mgmtsystem_review.review_report">
+        <t t-foreach="doc.line_ids" position="before">
+            <h2>
+                <strong>Satisfaction Survey Answers:</strong>
+            </h2>
+            <t t-foreach="doc.response_ids" t-as="response">
+                <h3>
+                    <strong>
+                        <span t-field="response.user_input_line_ids[0].value_text" />
+                        (<span t-field="response.user_input_line_ids[1].value_text" />)
```
carlos-lopez-tecnativa: When I try to print the report, I get this error:
image

mgmtsystem_review_survey/views/mgmtsystem_review_views.xml
```diff
@@ -0,0 +1,30 @@
...
+        <field name="arch" type="xml">
+            <page name="inputs" position="inside">
+                <separator string="Survey Answers" colspan="4" />
+                <field
+                    name="response_ids"
+                    colspan="4"
+                    nolabel="1"
+                    attrs="{'readonly':[('state','=','done')]}"
+                >
+                    <tree create="true">
```
carlos-lopez-tecnativa: I think this is not necessary; it's the default behavior.
```suggestion
                    <tree>
```

carlos-lopez-tecnativa (changes requested):

mgmtsystem_review_survey/report/review.xml
victoralmau: This was already happening before (currently in v15), but I have corrected it so that the correct data is displayed.

carlos-lopez-tecnativa (approved):
LGTM

pedrobaeza (approved):
/ocabot merge major

PR #694, opened 2025-07-23
[15.0][FIX] mgmtsystem*: Define sequence without company

victoralmau:

@Tecnativa


pedrobaeza (approved):
Yes, good default for sanity when creating other companies!

carlos-lopez-tecnativa (approved):

marcelsavegnago (approved):

PR #701, opened 2025-07-29
[16.0][FIX] mgmtsystem*: Define sequence without company

victoralmau:
FWP from 15.0: https://github.com/OCA/management-system/pull/694

@Tecnativa


carlos-lopez-tecnativa (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #703, opened 2025-07-29
[15.0][IMP] mgmtsystem: Define cumulative permissions

victoralmau:
Define cumulative permissions

**Before**
antes

**After**
despues

@Tecnativa


pedrobaeza (commented):
Have you checked that the permissions continue the same with this change?

victoralmau:
I haven't seen any problems with this, but of course, there are many modules that depend on it. 
Honestly, I find it strange that no one has noticed this until now (even in higher versions).

pedrobaeza (approved):

giarve (approved):

PR #707, opened 2025-07-30
[18.0][FIX] mgmtsystem*: Define sequence without company

victoralmau:
FWP from 17.0: https://github.com/OCA/management-system/pull/704

@Tecnativa


pedrobaeza (approved):

carlos-lopez-tecnativa (approved):
LGTM

PR #708, opened 2025-08-04
[17.0][IMP] mgmtsystem_review: Separate the survey part in a new module (mgmtsystem_review_survey)

victoralmau:
FWP from 16.0: https://github.com/OCA/management-system/pull/706

Separate the survey part in a new module (`mgmtsystem_review_survey`)

Before, if you needed to install `mgmtsystem_review` (because it was a dependency of `mgmtsystem_quality`) and you didn't use or want to use survey it was "mandatory", now, the new module `mgmtsystem_review_survey` adds the survey part to make that part optional.

Please @pedrobaeza and @carlos-lopez-tecnativa can you review it?

@Tecnativa TT57340

PR #715, opened 2025-09-22
[18.0][IMP] mgmtsystem_review: Separate the survey part in a new module (mgmtsystem_review_survey)

victoralmau:
FWP from 17.0: https://github.com/OCA/management-system/pull/708

Separate the survey part in a new module (`mgmtsystem_review_survey`)

Before, if you needed to install `mgmtsystem_review` (because it was a dependency of `mgmtsystem_quality`) and you didn't use or want to use survey it was "mandatory", now, the new module `mgmtsystem_review_survey` adds the survey part to make that part optional.

@Tecnativa TT57340


setup/mgmtsystem_review_survey/odoo/addons/mgmtsystem_review_survey
```diff
@@ -0,0 +1 @@
+../../../../mgmtsystem_review_survey
```
pedrobaeza: These files are not needed in this version.

pedrobaeza (approved):

pilarvargas-tecnativa (approved):

PR #741, opened 2025-09-26
[ADD] mgmtsystem_objective

etobella:
This module provides a way to define objectives.

TODO:

- [x] improve security
- [x] Add some tests

For the future:

- Create an automcompetion system
- Integrate with reviews

Before following the todo, it would be interesting to have some comments on the concept


etobella:
I added some JS Widget to make the experience smoother.

image

etobella:
It will fail because we need https://github.com/OCA/management-system/pull/742

luisDIXMIT (approved):
Code review and tested on runboat. LGTM!!

mgmtsystem_objective/models/mgmtsystem_indicator.py
```diff
@@ -0,0 +1,59 @@
...
+        [
+            ("draft", "Draft"),
+            ("active", "Active"),
+            ("inactive", "Inactive"),
+            ("cancelled", "Cancelled"),
+        ],
+        default="draft",
+        required=True,
+    )
+    value_ids = fields.One2many("mgmtsystem.indicator.value", "indicator_id")
```
FernandoRomera, line 36: Do you think we could indicate the type of unit of measurement for the values, such as liters, kilograms, degrees Celsius, etc.?

value_unit_id = fields.Many2one("uom.uom")

mgmtsystem_objective/models/mgmtsystem_indicator.py
```diff
@@ -0,0 +1,59 @@
...
+    _name = "mgmtsystem.indicator"
+    _description = "Indicator"
+    _inherit = ["mail.thread", "mail.activity.mixin"]
+
+    name = fields.Char(required=True)
+    objective_id = fields.Many2one("mgmtsystem.objective", required=True)
+    description = fields.Html()
+    frequency = fields.Selection(
+        [
+            ("monthly", "Monthly"),
```
FernandoRomera, line 18: daily and weekly?

mgmtsystem_objective/models/mgmtsystem_indicator.py
etobella: I thought about it and I had multiple doubts. There is for exemple percentages (growth...) and it is not clear when to use it. For this reason I avoided it

mgmtsystem_objective/models/mgmtsystem_indicator.py
etobella: for weekly I can find reasons (using the system for OKRs for example), but it is hard for me to justify daily. This indicators should be used in reviews. We don't do daily quality reviews usually. For that we have mgmt_system_indicators, that allows us to define indicators reports.

ValentinVinagre (approved):
🫶

etobella:
@FernandoRomera I added the weekly

If the unit is relevant, I can add it, but it will not be required :wink: WDYT?

FernandoRomera (approved):
LGTM

PR #745, opened 2025-10-02
[18.0][IMP] mgmtsystem_nonconformity: Improve view. Add tests

etobella:
I detected that this module was colliding with account_reconcile_oca. the chatter view there was not working. With this change, everything works better.

Also, I added a much cleaner patch to make it work. Now the code is simpler too (most of the changes comes from adding tests)

Also, I added some tests and improved the view:

Before:

image

Now:

image


pedrobaeza (approved):

luisDIXMIT (approved):
Code review and tested on runboat, LGTM!!

PR #749, opened 2025-10-02
[18.0][FIX] mgmtsystem: typo URL github

FernandoRomera:


FernandoRomera:
@dreispt 
Can you review this PR?

pedrobaeza (approved):
Although I still don't like the system of putting more modules to install in the base one.

/ocabot merge patch

dreispt:
@pedrobaeza Why would that be a concern?

pedrobaeza:
We have told several times about that: the modules are not migrated (I see the trick for that) or directly the modules are not included in the corresponding instance, provoking for the users (specially when there's an integrator in the middle) frustration about seeing an option not available.

PR #751, opened 2025-10-06
[18.0][ADD] mgmtsystem_review_objective

etobella:
integrates objective and reviews.

Some changes on the base modules where necessary to make it work perfectly.


ValentinVinagre (approved):
LGTM 👍🏻 
I would only change the start to end dates so that they can be indicated in one go and not by field, with a datapicker.

etobella:
Sure! 
image

luisDIXMIT (approved):
Tested on runboat and LGTM!

PR #753, opened 2025-10-10
[18.0][IMP] mgmtsystem*

etobella:
This is a Proof of Concept.

We are creating a new model to link documentation of the mgmtsystem. In 18, we are doing some hooks for making it work without extra modules, but in 19, we should split in 2 modules to make the functionality work better.

Todo:
- [ ] Apply the same logic to all other references to document.page
- [ ] To do migration scripts
- [x] Add some tests

Now, documents can have severel "types" and a kind (Manual, PRocedure...)

When clicking on them, Odoo is able to decide if it should open a Document page or this registry:

image

To make the navigation natural, I did the following in nonconformities:

image

(you need to press the button to open it)

@max3903 @ValentinVinagre can you make a functional test? Right now, I have only done nonconfomities.


max3903:
@etobella 

- [ ] There should be an option in the settings to install mgmtsystem_document
- [ ] I am not a big fan of the 2 buttons and the wizard. If you installed mgmtsystem_document to use page, you don't want link, so the possibility to create link should be hidden.
- [x] I created a page with the "Add Page" button, the template of the category was not copied to the content.
- [x] On a page, I clicked on "New", selected the category, entered the content and saved. My new procedure doesn't appear in the list.

etobella:
I made it work without this extra module to avoid breaking installations. The module should appear in 19.

I think both have sense together. Some departments might have their own way for documentation.

it is true about the new button. I will try to find an option for it 🤔

etobella:
two of your comments have been attended. For the first two, well, I think that the current approach is correct. If you want to avoid buttons, we can do a new desplegable button, but I would expect to have both systems active.

max3903:
@etobella 

I am not sure to follow: You either manage the content in Odoo or outside. You can't select both. One of the requirements of most standards is to centralize things.

etobella:
I disagree. systems ask for order, no centralization.

It is usual to have everything centralized, but it is not a requirement.

diggy128:
My 2 cents here.
True, ISO standards just set the requirement to create and control documents. How this is done and whether it is centralized or not is a matter of the implementation, not the standard.

This change (split) would make it cleaner for users who keep their documentation in another system.
However, I agree with @max3903 about the wizards, and furthermore, this should be transparent to users who opt to keep their documentation in odoo.

I can see the benefit of being able to use both link to external files AND a document.page. However, in a controlled environment, it is the difference of having a draft document in odoo with which other users can interact (primarily via the chatter) to a draft .doc file a user uploaded "in a hurry" breaking the document contribution and collaboration  procedure.

At an extra point, I would wonder: If document.page is not required, wouldn't it be possible to reuse ir.attachment with link type instead of creating a new model?

etobella:
just to be clear, I am not using files, I am personally using a git repo for the documentation for some reasons:

- I can use tools like sphinx to make it elegant and even add some extra functionalities like detecting KPIs and make a resume page
- It is easier to detect changes. In my experience using document.page is not optimal for diff, as odoo is cleaning it and it is hard to follow the real changes. In a case, the only complaint of external auditors is this issue
- The workflow for approval is quite easy for technical companies

I understand that the wizards are not optimal, but was a first solution and I can make improvements to avoid it. I was just showing the main concept here.

About using one or another, we can simplify it by setting the type as a model, making it archivable and avoiding future issues.

I know that it is not optimal right now, but I am talking about the concept mainly. If it is OK for you, I can proceed with these changes

diggy128:
> - I can use tools like sphinx to make it elegant and even add some extra functionalities like detecting KPIs and make a resume page

I've also somewhat tried to do this in odoo. So far I got to auto-create an index page (with versions, dates and authors) in each page category that works smoothly.

> - It is easier to detect changes. In my experience using document.page is not optimal for diff, as odoo is cleaning it and it is hard to follow the real changes. In a case, the only complaint of external auditors is this issue

I totally agree. I pushed a PR trying to make the diff easier to use but had mixed results. And Odoo writing to and reformatting the tags behind the scenes makes it almost useless.

> - The workflow for approval is quite easy for technical companies

Indeed this is a pain point in Odoo. 8/10 times for whatever reason, a new draft is created when a user opens a document.
 

How about mgmtsystem.document inheriting document.page via delegation (aka _inherits)?
I haven't used delegation since v8, so I don't know the current culprits but wouldn't that give extra flexibility and allow to handle document.page in its own workflow (if module is installed) without extra wizardry in mgmtsystem? Would a view override in the procedures page work to use the view for document.page?

P.S. We have been using Odoo for all documention and management system requirements since v8 in our company with high success BUT with equally high pain regarding document.page, so I strongly believe knowledge and mgmtsystem would benefit from any extra love :) I'll be available, should you want it, for any extra feedback, testing, etc.

And lastly, thank you for you efforts!

etobella:
_inherits is not an option, because that would mean that all mgmtsystem.document generate a document.page. The only option I find for this is to create a `mgmtsystem.document` class similar to `ir.actions` and then we have one subclass for type (reusing the table) that allows us to decide (there we can add the inherit). I am not 100% if that will work, but might have sense. We can give it a try, however, it will be hard to handle (one menu for each one and probably we need to specify a group to decide when to use pages, links or both)

Another option would be to embed the data there, but I am not 100% sure if that will work correctly.

About the diff on document pages. Have you ever thought of transforming the HTML to markdown and do the diff there?

Just to be clear, by using sphinx, we have several improvements, like cross references, using extensions to simplify everything (for example, you can use sphinx_needs) and generation to PDF natively by using Latex.

On a tech company, it is quite easy to use this kind of tools (more natural than document.pages). It is not for all the companies, but it has its target.

diggy128:
> _inherits is not an option, because that would mean that all mgmtsystem.document generate a document.page. 

Is this really the case even when no delegated fields are displayed on the form? Didn't remember that.

> The only option I find for this is to create a `mgmtsystem.document` class similar to `ir.actions` and then we have one subclass for type (reusing the table) that allows us to decide (there we can add the inherit). I am not 100% if that will work, but might have sense. We can give it a try, however, it will be hard to handle (one menu for each one and probably we need to specify a group to decide when to use pages, links or both)
> 
> Another option would be to embed the data there, but I am not 100% sure if that will work correctly.

I'm out of the office for a few days. I can study this option once I get back.

> About the diff on document pages. Have you ever thought of transforming the HTML to markdown and do the diff there?

In fact, I did at some point in v8 but didn't want to bring extra dependencies and don't have much experience with MD.

> Just to be clear, by using sphinx, we have several improvements, like cross references, using extensions to simplify everything (for example, you can use sphinx_needs) and generation to PDF natively by using Latex.
> 
> On a tech company, it is quite easy to use this kind of tools (more natural than document.pages). It is not for all the companies, but it has its target.

I fully understand and agree. Back in 2012, before we delved in odoo, we kept all ISO documentation and records in tiki wiki. I still miss some of its features, and continuous problems with html in odoo fields make me wonder sometimes if it's worth the pain to have a consolidated system.

etobella:
@max3903 @pedrobaeza your thoughs would be welcomed in this topic before digging more in it :smile:

etobella:
@OCA/management-systems-maintainers What do you think about this?

pedrobaeza:
I have tried in runboat and right now seems confusing how to link that documents.

etobella:
Ok, after giving a 2nd thought I think I can do the following to solve all the issues:

1. add a new module that allows to link document page to an external url (that solves the issue of external documents)
2. Add a new field on document.page to mark which docs are procedures, in order to remove the search in english.... https://github.com/OCA/management-system/blob/18.0/mgmtsystem_nonconformity/views/mgmtsystem_nonconformity.xml#L239
3. Propose a new diff way to make it work with html_2_markdown, diffs will be smaller:

image

Changes might be less aggressive this way

etobella:
For 2: https://github.com/OCA/knowledge/pull/563
For 3: https://github.com/OCA/management-system/pull/756

@pedrobaeza @max3903 Do you see it cleaner now?

pedrobaeza:
OK, this sounds better.

PR #756, opened 2025-10-27
[18.0][IMP] document_page_procedure: Remove dependency on name and make a computed field for this

etobella:
It is cleaner now.


document_page_procedure/__manifest__.py
```diff
@@ -3,13 +3,17 @@
 
 {
     "name": "Document Management - Wiki - Procedures",
-    "version": "18.0.1.0.1",
+    "version": "18.0.1.1.0",
```
max3903, line 6:
```suggestion
    "version": "18.0.2.0.0",
```

etobella: Done

etobella:
I made the same logic for manuals.

Only choice to do. What shall we do with enviromental aspects? Shall we create a new field just for that?

diggy128:
> I made the same logic for manuals.
> 
> ... What shall we do with enviromental aspects? Shall we create a new field just for that?

I think if we down this path, it is likely that many more fields like that will come up. Occupational Health and Safety if the first that comes to mind. 

Why don't we duplicate the picking type concept?

We define a document type selection field with document types: Document, Work Order, Procedure, Process, Manual
This field sets the document type hierarchy, allows to set domains and translate the selection values.
It also allows to add customization without adding extra fields.

Both menus can be set for each type and a generic one, much like the picking dashboard and menus for each picking type.

etobella:
Good for me! I will make the proposal @diggy128

etobella:
@diggy128 I finally had the time to do the promised work :wink:

diggy128:
Allow me a few thoughts:

1. Do we really need all these simple modules just to add a selection in a field? I think all, except mgmtsystem_nonconformity and mgmtsystem_audit would fit nicely in document_page_mgmtsystem and minimize registry pollution and module dependency with tiny modules.
2. I think mgmtsystem_page_type should only represent the document hierarchy and only have:
[("manual", "Manual"),
("process", "Process"),
("procedure", "Procedure"),
("work_instruction", "Work Instruction"),
]
Each one should have a menu under a submenu Management System / Documentation / Documents / ...
and Hazards should be Management System / Documentation / Hazards
This would separate plain written documentation from more "advanced" documentation objects.
3. Add an extra many2one field mgmtsystem_system_id in document.page that points to the system the document is linked.
4. We could add these two fields in the search panel as filters.
5. Organizational structure should be done using document category if needed (such as warehouse, production, office, etc)

So,
DATA:
```    
    <record id="mgmtsystem_system_environmental_aspect" model="mgmtsystem.system">
        <field name="name">Environmental Aspect</field>
    </record>
    <record id="mgmtsystem_system_health_safety" model="mgmtsystem.system">
        <field name="name">Health and Safety</field>
    </record>
```

- environmental_manual:
```
mgmtsystem_page_type = manual
mgmtsystem_system_id = env.ref(
        "mgmtsystem.mgmtsystem_system_environmental_aspect
    )
```

- health_safety_manual
```
mgmtsystem_page_type = manual
mgmtsystem_system_id = env.ref(
        "mgmtsystem.mgmtsystem_system_health_safety
    )
```

This could allow for better separation between document level and system the document applies to.
These fields could also be used to do a management system evaluation viewing document.page in a pivot view with page type on one axis and system or category on the other. This could show the average number of work instructions per procedure or point out procedures that have way many work instructions (possibly indicating that the procedure is not well defined, too complex or should be further split).

etobella:
One solution for that could be adding generic groups for enabling them. WDYT?

If we make a M2O field, we could activate and deactivate them, making it easier to handle this permissions. For 18 we might need to keep this module, but they would be removed in 19 (I would leave them empty for now)

max3903:
@etobella What was the issue with using category? How do we prevent a mismatch between category and type?

etobella:
The problem is translations.... if you use it in spanish, it does not work anymore.... the same happens if you change the name of the category to fit to your own definition.

max3903 (approved):

PR #757, opened 2025-11-02
[18.0][FIX] mgmtsystem_nonconformity: Add groups in fields to avoid issues. Refactor to avoid creation of fields in mail.thread

etobella:
Otherwise, errors might be risen when doing search.


pedrobaeza (commented):
Isn't there other possibility to not override this generic model? Otherwise, while exporting POT/PO files, all the models got contaminated with these 2 fields terms being exported.

etobella:
@pedrobaeza WDYT about the new change?

pedrobaeza (approved):
Yes, please!

luisDIXMIT (approved):
Code review and tested on runboat, LGTM!

PR #763, opened 2025-12-10
[18.0][MIG] mgmtsystem_security_event: Migration to 18.0

trixocom:
Hello, I migrated this module into 18.0
I'm submitting it for review. It works as far as I tested it.

Thanks
Hector Quiroz
Trixocom


etobella (changes requested):
A few things. It was a migration of which module?

I only found one with the same name on v8. If that is the case, you didn't follow the migration guide.

Also, use proper commit names please, leave the migration on a single commit (or adding if it is not a migration) and review pre-commit and tests.

trixocom-dev:
Sorry, it is the migration of mgmtsystem_security_event to v18.
there is no  mgmtsystem_security_event on v18 branch.

etobella:
In that case, you didn't follow the migration styles and your CI is failing.

trixocom-dev:
Ok Sorry, will try to fix it as soon as I get the time.

PR #764, opened 2026-01-02
[17.0][IMP] mgmtsystem: Define cumulative permissions

victoralmau:
FWP from 16.0: https://github.com/OCA/management-system/pull/762

Define cumulative permissions

**Before**
antes

**After**
despues

@Tecnativa


pedrobaeza (approved):

pedrobaeza:
@giarve as you review the previous patch, can you check the fw-port?

giarve (approved):
LGTM

PR #765, opened 2026-01-19
[18.0][IMP] mgmtsystem: Define cumulative permissions

victoralmau:
FWP from 17.0: https://github.com/OCA/management-system/pull/764

Define cumulative permissions

**Before**
antes

**After**
despues

@Tecnativa


pedrobaeza (approved):
Let's finish the forward port chain

/ocabot merge minor

PR #766, opened 2026-01-27
Wijzigingen Open2Bizz

dochse:


flotho (changes requested):
Hi, thanks for your PR
could you split the modules please.
No one will read all of this at once.

PR #769, opened 2026-02-22
[18.0] [FIX] management_system: Update kanban view

diggy128:
Kanban view classes changed in v18
- This greatly simplifies kanban view to use flex classes
- Adds a type-color map dict  that maps to color index from o-colors in secondary_variables.scss (This removes hard coded colors, promotes uniformity and allows to easily override if needed)
- Adds computed color field to replace (not working kanban_color) card hightlight color
- Removes deprecated classes

### DIFF
--- a/mgmtsystem_action/tests/test_create_action.py
+++ b/mgmtsystem_action/tests/test_create_action.py
@@ -47,6 +47,7 @@ class TestModelAction(common.TransactionCase):
             }
         )
         self.assertEqual(new_record.reference, self.env._("SampleReference"))
+        self.assertEqual(self.record.color, 1)
 
     def test_case_close(self):
         """Test object close state."""
--- a/mgmtsystem_action/models/mgmtsystem_action.py
+++ b/mgmtsystem_action/models/mgmtsystem_action.py
@@ -5,6 +5,13 @@ from datetime import datetime, timedelta
 
 from odoo import api, exceptions, fields, models
 
+TYPE_COLOR_MAP = {
+    "immediate": 1,
+    "correction": 2,
+    "prevention": 4,
+    "improvement": 10,
+}
+
 
 class MgmtsystemAction(models.Model):
     _name = "mgmtsystem.action"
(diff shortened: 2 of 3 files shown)


pedrobaeza:
Can you please put a before/after screenshots?

Please also check the pre-commit.

diggy128:
@pedrobaeza 
BEFORE:
before

AFTER:
after

~~Can't get codecov to cover the for loop given the existing test~~ Commit didn't pickup the test changes

mgmtsystem_action/tests/test_create_action.py
```diff
@@ -47,6 +49,8 @@ def test_create_action(self):
             }
         )
         self.assertEqual(new_record.reference, self.env._("SampleReference"))
+        self.record._compute_color()
```
pedrobaeza: You shouldn't need to call the compute method directly.

pedrobaeza (commented):
Please remove `[18.0]` from the commit message.

mgmtsystem_action/tests/test_create_action.py
```diff
@@ -47,6 +49,7 @@ def test_create_action(self):
             }
         )
         self.assertEqual(new_record.reference, self.env._("SampleReference"))
+        self.assertEqual(self.record.color, TYPE_COLOR_MAP[self.record.type_action])
```
pedrobaeza: Sorry for not saying before. On tests, you shouldn't replicate the compute logic, but to put the expected fixed data. Example: `self.assertEqual(self.record.color, 2)`

diggy128: Done. I hardcoded it in the beginning, but thought that a change in the map should also change the test, so I went with the dict lookup.

Thank you for the clarification.

pedrobaeza: Yeah, and in that case that the mapping changes, the test will highlight that "something has changed", and should be adapted accordingly.

pedrobaeza (approved):

victoralmau (approved):

PR #770, opened 2026-03-26
[16.0][IMP] mgmtsystem_evaluation: allow update feedback and note

luisDIXMIT:
Before, in an evaluation, when we chose a template and then changed it, the feedback and note fields weren’t updated, even if we cleared their content, because the HTML fields kept `<p><br/></p>`.

### DIFF
--- a/mgmtsystem_evaluation/models/mgmtsystem_evaluation.py
+++ b/mgmtsystem_evaluation/models/mgmtsystem_evaluation.py
@@ -138,9 +138,9 @@ class MgmtsystemEvaluation(models.Model):
                 record.res_id = False
                 record.model = template.model
                 record.model_id = template.model_id
-            if not record.feedback:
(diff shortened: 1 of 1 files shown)


max3903:
@luisDIXMIT I get your point with the `<p><br/></p>` but are you fine with the following scenario: 

1. User sets the feedbacks and notes
2. User changes the template --> Feedbacks and notes are lost as they get reset to the template value.

Maybe we could have: `if record.feedback == '<p><br/></p>'`

luisDIXMIT:
> @luisDIXMIT I get your point with the `<p><br/></p>` but are you fine with the following scenario:
> 
> 1. User sets the feedbacks and notes
> 2. User changes the template --> Feedbacks and notes are lost as they get reset to the template value.
> 
> Maybe we could have: `if record.feedback == '<p><br/></p>'`

Thanks for your review @max3903 now it is better, thanks!

Issue #786, opened 2026-05-09
mgmtsystem_nonconformity_*: refactor to mgmtsystem_event


diggy128:
I would like to propose a discussion regarding the functional scope and naming of the mgmtsystem_nonconformity module.

**Is your feature request related to a problem?**

In practice, many organizations implementing integrated management systems (IMS) use the same operational workflow not only for nonconformities, but also for:

- change requests
- improvements
- preventive actions
- risks
- incidents
- audit findings
- supplier issues
- customer complaints
- food safety changes
- business continuity changes
- management-of-change processes

The current mgmtsystem_nonconformity workflow already covers most of the lifecycle required for these records:

- submission
- assessment
- approval
- action assignment
- implementation
- verification
- effectiveness review
- closure

As a result, mgmtsystem.nonconformity could be reused as a generalized “managed change/event” object.

However, the current terminology (“nonconformity”) is semantically restrictive and does not accurately represent the broader ISO use cases already covered by the workflow.

**Describe the solution you'd like**

Modern integrated management systems are increasingly centered around:

- risk-based thinking
- change management
- organizational governance
- preventive management
- continuous improvement

rather than purely corrective handling of failures.

The existing workflow structure in mgmtsystem_nonconformity is already very close to a generalized ISO Management of Change (MoC) process.

A refactoring renaming mgmtsystem.nonconformity to mgmtsystem.change_request along with adding a change_type field (having nonformormity + the above reasons) would:

- cover all the above cases
- generalize / standardize the workflow for change management throughout the organization regardless of sector (e.g. manufacturing or services)
- handle easier any future changes or adaptations
- provide a true 360° overview of changes in the organization. This could also be applied beyond the boundaries of ISO and standards implementation (i.e. affecting the management systems or otherwise).

**Benefits**
- Better alignment with modern ISO integrated systems
- Reduced duplication of custom modules
- Easier adoption for MoC workflows
- Better semantics for users and auditors
- More natural support for ISO 22000 / 45001 / 22301
- Stronger support for integrated management systems (IMS)

This change could also skyrocket participation from employees:
- creating a change request if much more relatable for employees of any level (in comparison to non conformity)
- can provide a KPI for process / W.O. effectiveness / efficiency
- can provide a KPI for employees' engagement to continuous improvement
- can provide a KPI for employees' understanding of the requirements of M.S.

**Describe alternatives you've considered**
New mgmtsystem_change module

**Additional context**
**ISO standards explicitly requiring controlled changes**

[...]

PR #787, opened 2026-05-17
[18.0][ADD] mgmtsystem_event: New Module

diggy128:
- Copy of mgmtsystem_nonconfirmity
- Remove i18n directory
- Rename all nonconformity -> event
- Incorporate changes by maintainer-tools

This is the starting point of refactoring mgmtsystem_nonconformity as talked about at #786


diggy128:
@etobella @mymage This is the first batch of changes.
Since I have a very volatile schedule, the changes will come in random order. I'll commit often, keeping the module in a working state so that I can get your feedback. 
Once we are done, I'll squash the commits.

Next up is changing state in stage_id. As is now, It is of no use as it is a one-to-one duplicate of stage_id data.
I'll change state and keep: draft, open, closed

diggy128:
Currently Stage_id and state have the same values.
Stage_id is used for the kanban view and state for the functional part of the form view (readonly and invisible).
Is that separation needed?

I am thinking of merging the two and keep only state.

What do you think @max3903 @etobella @mymage ?

etobella:
You probably need states, because it is the way to know if a field should appear or not. stage tables is interesting for  allowing users to change the behavior, but we need to force somehow the appearance of fields.

mymage:
Hi all, I will try to share my opinion but be aware of my imperfect English and coding limitation.

Event_type in event: for me "immediate" (something related to time) is not coherent with other terms (related to the reason). What do you think to use 'fixing' or 'repairing'?
Other question, do you think that will be better to let user define his own types? Because in the origin discussion I used other values as "near miss", "nonconformity", ...

Stage and State. 
I agree to @etobella note, my only doubt (because i don't know all management system types) is if all system managed is covered by the actual states defined.
For stages if I understand well, when I define some stages these are available on every system managed.
If so I'm not sure that if I define two different management system in a company, both must have the same stages.

PS: curiosity: I tried the green runboat but it doesn't work, why?

diggy128:
> PS: curiosity: I tried the green runboat but it doesn't work, why?

@mymage  Runboat expired. It should work now

PR #789, opened 2026-05-20
[18.0][MIG] mgmtsystem_environment: migrate environment and health_safety to 18.0

jbaldellouelepe:


InformatiqueProg:
@jbaldellouelepe , @max3903 As there is already an ‘approved’ PR for the ‘mgmtsystem_health_safety’ migration, and it is preferable to handle each module in a dedicated PR.
Would it be a problem if I proposed migrating `mgmtsystem_environment` to v17 and then to v18? Or would you prefer to take care of it yourselves?

Issue #792, opened 2026-05-22
Module mgmtsystem_health_safety_manual is not available


InformatiqueProg:
## Module
mgmtsystem_health_safety_manual

## Describe the bug
Impossible to activate

## To Reproduce
**Affected versions**:

v17 and v18:

1.     When you go in Settings / Management Systems
2.     You want to activate => Health & Safety Manual
Provide a Health & Safety Manual template based on the ISO 45001 / OHSAS 18001 standard
3.     You have the following error

```
Invalid Operation

The following modules are not available: mgmtsystem_health_safety_manual
```

**Expected behavior**
That he activates

<img width="2560" height="1528" alt="Image" src="https://github.com/user-attachments/assets/a14d9155-8b3f-4fab-a668-9e5b1d444a46" />

max3903:
#799

Issue #788, opened 2026-05-20
Module `mgmtsystem_health_safety` is not available


InformatiqueProg:
## Module
mgmtsystem_health_safety

## Describe the bug
Impossible to activate

## To Reproduce
v17 and v18:

1. When you go in Settings / Management Systems
2. You want to activate => Hygiene & Safety
Manage your health and safety management systems (ISO 45001, OHSAS 18001).
3. You have the following error 
```
Invalid Operation
The following modules are not available: mgmtsystem_health_safety
```

**Expected behavior**
That he activates

**Additional context**
The `mgmtsystem_health_safety`  disappears with version 17.0 and i don't understand why...

<img width="2560" height="1528" alt="Image" src="https://github.com/user-attachments/assets/5bb2d32c-75b8-4703-81f4-12fd5a2979e6" />

max3903:
@InformatiqueProg Module needs to be migrated: #554 #600

max3903:
#794

InformatiqueProg:
Thanks, @max3903 , for migrating the `mgmtsystem_health_safety` module to V17.
Sorry, but is there a reason why this PR wasn't also ported to V18?
Would you prefer to do it? Otherwise, I can give it a try.

max3903:
@InformatiqueProg I have limited resources and availability. If you can do it, let me know.

InformatiqueProg:
I completely understand @max3903 , no problem.
I'll try to get that done today or tomorrow.

InformatiqueProg:
@max3903 Can you help for this question?
 I couldn't find any information on the wiki regarding this question.
Will a bot update the ".po/.pot" files once it's merged, or do I need to handle that in the commits for this migration?
FYI i have already done this => https://github.com/InformatiqueProg/management-system/commits/18.0-mig-mgmtsystem_health_safety/

max3903:
@InformatiqueProg Yes. The bot takes care of updating the translation template.

InformatiqueProg:
@max3903 Apparently I managed to pass the checks. ;-)
#840

PR #793, opened 2026-05-22
[18.0] fix mgmtsystem health safety

InformatiqueProg:
Solve #788 and #792


max3903:
@InformatiqueProg You need to separate branches to have one per module and follow the instructions here: https://github.com/OCA/maintainer-tools/wiki

PR #794, opened 2026-05-31
[17.0][MIG] mgmtsystem_health_safety: Migration to 17.0

max3903:
Migration of `mgmtsystem_health_safety` to Odoo 17.0 following OCA maintainer-tools guidelines.

## Changes applied
- Bumped version to `17.0.1.0.0`
- Ran pre-commit formatters (ruff, ruff-format — all passed with no changes)
- No Python models or XML views requiring framework changes (data-only module)

## Checklist
- [x] Tests pass
- [x] pre-commit passes
- [x] README updated if needed


max3903:
@InformatiqueProg Can you review this PR?

mgmtsystem/models/res_config.py
```diff
@@ -66,12 +66,12 @@ class MgmtsystemConfigSettings(models.TransientModel):
         help="Provide an environment manual template.\n"
         "- This installs the module mgmtsystem_environment_manual.",
     )
-    module_mgmtsystem_health_safety_manual = fields.Boolean(
+    module_document_page_health_safety_manual = fields.Boolean(
         "Health & Safety Manual Template",
         help="Provide a health and safety manual template.\n"
         "- This installs the module mgmtsystem_health_safety_manual.",
```
InformatiqueProg: @max3903 I think that the name in help text need to be changed too

mgmtsystem/models/res_config.py
```diff
@@ -66,12 +66,12 @@ class MgmtsystemConfigSettings(models.TransientModel):
...
     )
-    module_information_security_manual = fields.Boolean(
+    module_document_page_information_security_manual = fields.Boolean(
```
InformatiqueProg: @max3903 Good catch, I hadn't seen it before.
a) mgmtsystem_information_security module is missing 
b) wrong name for `module_information_security_manual`
Your right the good name would be `module_document_page_information_security_manual`, but actually the module name is `mgmtsystem_info_security_manual)`.
Module folder and files need to be renamed too in this case.

mgmtsystem/models/res_config.py
```diff
@@ -66,12 +66,12 @@ class MgmtsystemConfigSettings(models.TransientModel):
...
         "Information Security Manual Template",
         help="Provide an information security manual.\n"
         "- This installs the module information_security_manual.",
```
InformatiqueProg: @max3903 I think that the name in help text need to be changed too

InformatiqueProg (changes requested):
Thank you for correcting my bad work.
But i found some problems in your PR

max3903:
@InformatiqueProg Can you update your review?

mgmtsystem/models/res_config.py
InformatiqueProg: @max3903 I agree, this change needs to be made too, but I don't see in this PR the renaming of the "mgmtsystem_info_security_manual" folder to "document_page_information_security_manual"
it's normal ?

InformatiqueProg (changes requested):
Unless I'm mistaken, 
I think the renaming of the "mgmtsystem_info_security_manual" folder to "document_page_information_security_manual" is missing from this PR.

marcelsavegnago (approved):
LGTM

max3903:
> Unless I'm mistaken, I think the renaming of the "mgmtsystem_info_security_manual" folder to "document_page_information_security_manual" is missing from this PR.

I will revert the changes and focus this PR on the migration. The renaming will need to happen in a separate PR.

PR #828, opened 2026-06-20
[IMP] document_page_procedure: Add Gray Matter Logic as author, fix category and update icon

max3903:
This PR improves the `document_page_procedure` module for Odoo 19.0.


etobella:
@max3903 I am quite worried that you made all the migration without waiting for a single review.

In less than one hour you made the migration and committed without waiting for other maintainers to review. This is not they way to handle things

etobella:
Even worse, you didn't collaborate on this module, why you added your company as Author?

PR #835, opened 2026-06-20
[IMP] mgmtsystem_manual: Add Gray Matter Logic as author, fix category and update icon

max3903:
This PR improves the `mgmtsystem_manual` module for Odoo 19.0.


pedrobaeza:
This PR should be reverted as well, as:

- You haven't contributed to this module as can be seen in the commit history.
- The commit is put as migration to 19, while the migration was already done.
- Self-merged in an hour.

PR #842, opened 2026-06-29
[17.0][FIX] mgmtsystem_nonconformity: use correct Odoo 17 kanban state widget

miquelrosell99:
The widget `kanban_state_selection` does not exist in Odoo 17; it was renamed to `state_selection`. Because the old widget name is unknown, the `kanban_state` field on nonconformities falls back to a plain selection dropdown.

This PR updates both the kanban card and the form view to use the correct `state_selection` widget.

### DIFF
--- a/mgmtsystem_nonconformity/views/mgmtsystem_nonconformity.xml
+++ b/mgmtsystem_nonconformity/views/mgmtsystem_nonconformity.xml
@@ -185,7 +185,7 @@
                                 <div class="oe_kanban_bottom_right">
                                     <field
                                         name="kanban_state"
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Clear patch, so fast-tracking it:

/ocabot merge patch

Please fw-port it to 18

miquelrosell99:
Forward-port to 18.0 is not needed: the 18.0 migration of `mgmtsystem_nonconformity` already uses the correct `state_selection` widget for `kanban_state`, so this fix is already present in the 18.0 branch.

pedrobaeza:
Great, thanks for checking.

PR #845, opened 2026-06-30
[17.0][IMP] mgmtsystem_audit: make state statusbar clickable

miquelrosell99:
Allows users to change the audit state directly from the form view statusbar instead of relying only on the Close button.


miquelrosell99:
Replaced by #847 which adds an explicit Open button instead of making the statusbar clickable.

PR #854, opened 2026-07-06
[19.0][MIG] document_page_mgmtsystem: Migration to 19.0

etobella:
Applies changes from #756 to 19.0

I added a smart detection of migration to allow migration from 18 to 19 without executing the scripts.

@FernandoRomera @max3903


max3903:
@etobella I created various manuals (other than quality) and they don't show up in the Manuals menu.

max3903 (changes requested):
The Quality Manual category shows up in the Manuals menu.

max3903:
@etobella Both issues remain:

- [x] The Quality Manual category shows up in the Manuals menu.
- [ ] The Environment Manual don't show up in the Manuals menu.
- [ ] Go to Documentation > Manuals. Create a new manual. Select a different category (Environment Manual for example). The content remains with the Quality one.

etobella:
@max3903 I think both problems are solved

max3903:
@etobella Still some issues: https://github.com/OCA/management-system/pull/854#issuecomment-5023333760

etobella:
Can you give the steps that you did in runboat to have the issues? A video might work too

max3903:
@etobella 

* Go to the Runboat instance: http://oca-management-system-19-0-pr854-97c734f3306c.runboat.odoo-community.org
* Go to Management Systems > Documentation > Manuals
Screenshot from 2026-07-23 07-27-13

* Go to Knowledge > Pages
Screenshot from 2026-07-23 07-27-41

The new issue is related to the category being set by default when creating a new manual and the fact that the content is not updated when the category is changed (missing onchange).

etobella:
For me the problem is different.

image

You are making Environmental Manuals, that are not Manuals (Saved for Quality Manuals), are "Enviromental Manuals". There should be a specific menu for that.

Is that ok for you?

max3903:
@etobella The Manuals menu opens all the manuals. Search panel allows to filter based on the Quality, Environment, etc.

Having separate menu items for each does not make sense. Most organizations would have only one record in each menu.

etobella:
In this case, we should think about if we really nead enviromental Manuals category at all.

etobella:
If we want them to be shown as Quality manuals, we just need to remove environment_manual and use `quality_manual`as category. My recomendation would be to remove the module in 20.0 as it is not really a requirement anymore.

max3903:
@etobella We need it to provide the structure of the environment manual from ISO 14001. Categories are templates and hold the default content for the pages.

etobella:
Yes, but with the new change, you can do it yourself without using a module to configure it. For this reason I am saying that I would recommend to remove it in 20. Obviously, I will not recommend that approach in 19.

max3903:
@etobella I am not sure users will prefer entering all the content. I think they will appreciate having the template being provided.

etobella:
Well, that is something that we can decide from a PSC perspective on the migration. IMO, it would be better to have templates to upload on the system and not depend on unnecessary modules. At the end, we are just pushing data and in my experience, we are just adding a few categories and every company might have their own way for doing that.

PR #761, opened 2025-12-04
[16.0][FIX] mgmtsystem_nonconformity_quality_control_oca: fix opening client action permissions

giarve:

PR #862, opened 2026-09-10
[18.0][MIG] mgmtsystem_nonconformity_quality_control_oca: Migration to 18.0

anusriNPS:
Migrated mgmt_nonconformity_quality_control_oca to 18.0


SirPyTech (changes requested):
The module exists in `17.0` (https://github.com/OCA/management-system/tree/16c6976885b91539edc063d45dd54d04e06f713e/mgmtsystem_nonconformity_quality_control_oca), why has it been migrated from `16.0`?

anusriNPS:
> The module exists in `17.0` (https://github.com/OCA/management-system/tree/16c6976885b91539edc063d45dd54d04e06f713e/mgmtsystem_nonconformity_quality_control_oca), why has it been migrated from `16.0`?

Apologies, did not check version 17.0 related to the module. Hence, missed to migrate from it. Updated migration commit accordingly

mgmtsystem_nonconformity_quality_control_oca/views/qc_inspection.xml
```diff
@@ -49,7 +49,7 @@
     <!--    <record model="ir.ui.view" id="qc_inspection_tree_view">-->
     <!--        <field-->
     <!--            name="name"-->
-    <!--        >qc.inspection.tree (in mgmtsystem_nonconformity_quality_control_oca)</field>-->
```
SirPyTech: **suggestion**: We might remove all this commented code

mgmtsystem_nonconformity_quality_control_oca/models/qc_inspection.py
```diff
@@ -0,0 +1,49 @@
...
+
+    @api.depends("mgmtsystem_nonconformity_ids")
+    def _compute_mgmtsystem_nonconformity_count(self):
+        for rec in self:
+            rec.mgmtsystem_nonconformity_count = len(rec.mgmtsystem_nonconformity_ids)
+
+    def action_view_nonconformities(self):
+        action = self.env.ref(
+            "mgmtsystem_nonconformity.open_mgmtsystem_nonconformity_list"
+        ).read()[0]
```
SirPyTech: **suggestion**: `_for_xml_id` (https://github.com/odoo/odoo/blob/840e84e9ca85e1395659709dadacbce7f16a61e1/odoo/addons/base/models/ir_actions.py#L218) does the same thing and might prevent permission errors

SirPyTech (approved):
Thanks for the PR!
I found the following **issue**:
1. Login with Admin
2. Create a Nonconformity for an Inspection
4. Assign the groups Quality Control / Manager and Management System / Manager to Demo user
5. Login with Demo
6. In the inspection, click on the "Nonconformity" button

**Current behavior**:
Access error:
Image

**Expected behavior**:
See the records

This probably happens in other versions too, so please fix it in a new commit so it can be easily ported, the comment about `_for_xml_id` might fix it.
If you could also add a test that would be perfect 🤩

I am not blocking the migration though because does not depend on the migration changes.

**praise**: Nice work squashing the commits! 🙏

anusriNPS:
> Thanks for the PR! I found the following **issue**:
> 
>     1. Login with Admin
> 
>     2. Create a Nonconformity for an Inspection
> 
>     3. Assign the groups Quality Control / Manager and Management System / Manager to Demo user
> 
>     4. Login with Demo
> 
>     5. In the inspection, click on the "Nonconformity" button
> 
> 
> **Current behavior**: Access error: Image
> 
> **Expected behavior**: See the records
> 
> This probably happens in other versions too, so please fix it in a new commit so it can be easily ported, the comment about `_for_xml_id` might fix it. If you could also add a test that would be perfect 🤩
> 
> I am not blocking the migration though because does not depend on the migration changes.
> 
> **praise**: Nice work squashing the commits! 🙏

Thank you catching the error and `_for_xml_id` works as suggested. Included testcase as suggested.

However, in 16.0 a fix has been merged for same issue #761 so instead of modifying changes here in migration commit, ported the fix from 16.0 in #863 and included required testcase in a seperate commit.

sergiocorato (approved):

